PR #12, opened 2023-09-18
add basic skaffolding

bpmct:
fixes https://github.com/coder/testing-modules/issues/3.

I think we may outgrow this format as we make 5-10 modules, but this is a start


matifali (approved):
Great.

matifali (changes requested):

kylecarbs:
I guess this was closed because I force-pushed, this is my bad :(

PR #18, opened 2023-09-20
add gcp regions module

### LINKED ISSUE #8: GCP Regions/Zones Module
Figuring out all of the flag emojis and friendly names for GCP regions is difficult. Let's make a module that has all of these 

- [ ] Output with local of all modules
- [x] `coder_parameter` 

## Should have

- [x] User can specify via a module variable a subset of regions they want to support `["us-east-1", "us-east-2"]`
- [ ] User can disable the parameter and just consume the local? idk why they would want this but why not

matifali:
closes #8


stirby (approved):
This all looks great to me, do we intend to add the optional name/icon overrides to every region as they appear in AWS? I think it's a nice touch.

matifali:
Let me see if I can add that now. Otherwise, I can add in a separate PR.

PR #21, opened 2023-09-20
fix: gcp regions module

matifali:
fixes some issues and improves documentaion


stirby (approved):
Looks great, good call on the characters corresponding to region codes in each name.

PR #26, opened 2023-09-21
fix: azure regions 

stirby:
Added missing Azure regions and flags, updated errors in readme examples.


matifali (approved):

matifali:
@bpmct use squash and merge. This resulted in merging each commit individually.

bpmct:
oops

PR #36, opened 2023-09-24
chore: fix new.sh

bpmct:


.sample/README.md
```diff
@@ -9,9 +9,9 @@ tags: [community]
 
 # MODULE_NAME
 
-<-- Describes what this module does -->
+<!-- Describes what this module does -->
```
matifali, line 12: Have you verified that this is valid markdown comment now?
I think we need three dashes instead of 2.

.sample/README.md
```diff
@@ -9,9 +9,9 @@ tags: [community]
...
 
-<-- Add a screencast or screenshot here -->
+<!-- Add a screencast or screenshot here -->
```
matifali, line 14: Same here

matifali (approved):
LGTM. Just verify that the markdown comment is valid.

PR #39, opened 2023-09-24
add JupyterLab module

### LINKED ISSUE #9: Jupyter
## Must have 

- [ ] Auto-installs Jupyter with pip 
- [ ] Starts JupyterLab
- [ ] `coder_app`
- [ ] `coder_script`

## Should have

- [ ] Variable to specify version
- [ ] Variable to specify whether they want hub or lab

matifali:
closes #9


.sample/run.sh
matifali, line 1: chmod +x

new.sh
matifali, line 1: we should explicitly `chmod +x` the `run.sh`

.images/jupyterlab.webp
matifali, line 1: Let us put all screenshots in .images so that when someone uses the module, they don't pull the images.

bpmct: totally agree! was going to mention this and forgot

bpmct (approved):

PR #42, opened 2023-09-25
fix: update tags and fix module bugs, rewrite vscode-server

matifali:
1. I am adding both web and ide to the web ides and only ide to desktop ones like gateway and VS Code Desktop.
2. I added `parameter` to all that display parameters
3. I added `regions` to all three region modules
4. I added `jupyter` to  JupyterLab
5. I removed _jupyerhub_ module as it is not an IDE but a scalable way using K8s to host JupyterLab for teams.  We can do The Littlest JupyterHub, which also is more suitable for an example template than a module. Alternatively, we can do the [legacy Jupyter Notebook](https://github.com/coder/modules/issues/43).
6. Fixed a lot of random cosmetic and functional bugs


stirby (approved):
Tag updates look good, vscode-web -> vscode-server looks good.

PR #51, opened 2023-09-26
feat: add linting

kylecarbs:


matifali (approved):

matifali:
@kylecarbs it looks like most of the modules have example code blocks but linter is complaining.
Do They need to be before ##Examples heading?

matifali:
@kylecarbs our linter is now happy :D

PR #52, opened 2023-09-26
add fly.io region module

### LINKED ISSUE #27: Fly.io regions
- [ ] Test
- [ ] Add default
- [ ] Allow choosing a subset of regions

stirby:
Added fly.io region module to close #27. Needs updated README


fly-region/main.tf
```diff
@@ -0,0 +1,287 @@
...
+}
+
+variable "description" {
+    default = "The region to deploy workspace infrastructure."
+    description = "The description of the parameter."
+    type = string
+}
+
+variable "default" {
+    default = "ams"
```
matifali, line 25: Can you make the default value of default null and use a condition in coder_paramter to set default as null if var.default is null

matifali (approved):
Great work. Perfect examples and lovely screenshots.

fly-region/main.tf
stirby: Can do, the coder parameter default should already host the logic to do so. It checks if the default is in the filtered regions.

PR #60, opened 2023-09-28
Make `code-server` display name variable

TheZoker:
Currently the display name for the `code-server` application is just `code-server`, which personally I want to change to `VS Code`. This PR adds the option to set a custom display-name


matifali (approved):
Thanks

Issue #66, opened 2023-09-29
Ensure that not setting default or setting default to "" or `null` sets the default to `null` for all coder_parameter modules 


matifali:
While testing I noticed that

- [ ] JetBrains Gateway module fails if no default is set.
- [x] Azure regions module sets the default region to `eastus` if no default is set.
- [x] AWS regions module sets the default region to `us-east-1` if no default is set. 

Ideal behavior should be that setting the default to `null` should set the default of `coder_paramter` to `null` so that the user is forced to choose a value from the list.

stirby:
I notice when the default is `null` on a coder_parameter, the create workspace page reloads when no option is selected rather than throwing an error message. Is this the expected behavior?

matifali:
@stirby for parameters that have options when the default is `null`, it should ask the user to choose one of the options.

stirby:
Please see PR #72 for approval of `jetbrains-gateway` implementation.

PR #72, opened 2023-10-02
chore: allowed empty default for modules

stirby:
From #66, also added testing for `jetbrains-gateway`.

## Regions
Azure and AWS were simple, but there is no error message when no region is selected:

https://github.com/coder/modules/assets/58410745/9a9e317b-7764-427f-b298-86313de5fb33

Open to feedback on this.

## Jetbrains Gateway
The Jetbrains gateway default IDE satisfies a ton of logic in the template, so allowing `null` would lead to a lot of patchwork `try` statements. Now it's implemented to use the first IDE in the `jetbrains_ides` as the coder_parameter default for cleanliness. 

Let me know your thoughts.

### DIFF
--- a/aws-region/main.test.ts
+++ b/aws-region/main.test.ts
@@ -13,7 +13,7 @@ describe("aws-region", async () => {
 
   it("default output", async () => {
     const state = await runTerraformApply(import.meta.dir, {});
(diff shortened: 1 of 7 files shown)


matifali (approved):
Thanks. This is a good approach given the complexity of this module.

Issue #76, opened 2023-10-04
Filebrowser DB Location


djarbz:
I am looking for a way to modify where the filebrowser.db is stored.
Currently, it looks like it is saving in my working dir.

I would like it to save in `~/.config/...` but it is saving as `/workspace/repo/filebrowser.db`, which is my `coder_agent.dir`.

I think this should be the relevant configuration from my template for this issue.
```
locals {
  folder_name = try(trimsuffix(element(split("/", data.coder_parameter.repo.value), length(split("/", data.coder_parameter.repo.value)) - 1), ".git"), "")
  repo_owner_name = try(element(split("/", data.coder_parameter.repo.value), length(split("/", data.coder_parameter.repo.value)) - 2), "")
  repo_host = try(one(regex("^(?:https?:\\/\\/)?(?:[^@\\/\\n]+@)?(?:www\\.)?([^:\\/\\n]+)", data.coder_parameter.repo.value)), "")
  root_dir = "/workspace"
  project_dir = "${local.root_dir}/${local.folder_name}"
}

resource "coder_agent" "main" {
  arch                   = data.coder_provisioner.me.arch
  os                     = data.coder_provisioner.me.os
  dir                    = local.project_dir
  startup_script_timeout = 180
}

# A file browser for your workspace
module "filebrowser" {
  source = "https://registry.coder.com/modules/filebrowser"
  agent_id = coder_agent.main.id
  folder = local.root_dir
}
```

stirby:
I'll take a look into this and see if we can parameterize the `filebrowser.db` save location.

matifali:
Fixed in #78

PR #78, opened 2023-10-05
feat: add database_path specification to filebrowser

stirby:
Added the option to specify the save location for `filebrowser.db` in the `filebrowser` module much like the [database CLI flag](https://filebrowser.org/cli/filebrowser) in Filebrowser. 

Also added testing for this module. 

In response to #76.


matifali (approved):
Nice.

PR #80, opened 2023-10-09
fix(git-clone): change to work with volumes fixes #79

### LINKED ISSUE #79: [git-clone] - doesn't work with volumes (in k8s)
Discussion started here: https://discord.com/channels/747933592273027093/1157627766607454288

Basically, the git-clone module doesn't take into account that a volume might have already generated the directory path and with such a volume in place, the cloning of the repo will never happen. 

The advantages with using a volume are a workspace (container) can be stopped and restarted ( to be updated or to save resources or just to get a fresh container running) and the dev can basically continue where she left off. Without a volume, the cloning, installing, building, etc. etc. all has to happen again, before the dev can continue. Also, if the dev had made changes to the repo and didn't push them to origin, before restarting the workspace, all changes would be lost. 

I'll put in a PR to fix. 

Scott

smolinari:
Hi,

fixes #79 

I hope this is ok. It's tested and works in my k8s setup (the shell script that is).

Please note, `git clone` also checks for an empty folder and not if the folder exists. So, the logic is now the same to match git clone's. :)

Scott

### DIFF
--- a/git-clone/run.sh
+++ b/git-clone/run.sh
@@ -23,7 +23,7 @@ if ! command -v git >/dev/null; then
   exit 1
 fi
 
-# Check if the director for the cloning exists
+# Check if the directory for the cloning exists
 # and if not, create it
 if [ ! -d "$CLONE_PATH" ]; then
   echo "Creating directory $CLONE_PATH..."
--- a/git-clone/run.sh
+++ b/git-clone/run.sh
@@ -23,16 +23,19 @@ if ! command -v git >/dev/null; then
   exit 1
 fi
 
-# Check if the directory exists...
+# Check if the director for the cloning exists
+# and if not, create it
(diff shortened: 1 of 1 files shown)


matifali:
Thanks. @smolinari, can you also add a test [here](https://github.com/smolinari/coder-modules/blob/main/git-clone/main.test.ts) for this use case?

git-clone/run.sh
```diff
@@ -23,16 +23,19 @@ if ! command -v git >/dev/null; then
   exit 1
 fi
 
-# Check if the directory exists...
+# Check if the director for the cloning exists
```
kylecarbs, line 26:
```suggestion
# Check if the directory for the cloning exists
```

kylecarbs (approved):

PR #86, opened 2023-10-11
slackme: exit with command code and quote command

kylecarbs:


slackme/slackme.sh
```diff
@@ -85,3 +86,5 @@ SLACK_MESSAGE=$(echo "$SLACK_MESSAGE" | sed "s|\\$DURATION|$PRETTY_DURATION|g")
 curl --silent -o /dev/null --header "Authorization: Bearer $BOT_TOKEN" \
     -G --data-urlencode "text=$${SLACK_MESSAGE}" \
     "$SLACK_URL/api/chat.postMessage?channel=$USER_ID&pretty=1"
+
+exit $CODE
```
mafredri, line 90: The `set -e` above may still exit early, so we may mask the command exit code.

We also probably need `"` in `COMMAND="$(echo "$@")"`, but this may include a newline, so I'd suggest `COMMAND="$*"`.

mafredri (approved):

matifali:
@kylecarbs @bpmct, do we still need this in?

matifali:
closing due to inactivity

PR #87, opened 2023-10-12
Add Exoscale zone Module

masterwendu:
Like Google, AWS and Azure regions, this module provides predifined human readable exoscale cloud zones.

I built it up like the region modules are build.

I wasn't sure if I can set verified to `true` so I let it false. Please tell me what to do to verify a module.


masterwendu:
BTW I use the description in the parameter as it was intruduce to be displayed in v2.3.0

stirby (approved):

stirby:
You'll need to run `bun fmt` to meet the formatting CI requirements and merge.

PR #91, opened 2023-10-16
feat: share variable

ericpaulsen:
adds support for the `share` argument to be set at the module level, for applications that folks might want to share.


matifali (approved):

matifali:
@ericpaulsen you need to run `bun fmt`

PR #94, opened 2023-10-20
feat: add git-commit-signing module

phorcys420:
There are no tests since it would require a valid Coder account.

https://github.com/coder/coder/issues/8922
cc: @matifali

### DIFF
--- a/git-commit-signing/README.md
+++ b/git-commit-signing/README.md
@@ -1,0 +1,23 @@
+---
+display_name: Git commit signing
+description: Configures Git to sign commits using your Coder SSH key
+icon:  ../.icons/git.svg
+maintainer_github: phorcys420
+verified: false
+tags: [helper, git]
+---
+
+# git-commit-signing
+
+This module downloads your SSH key from Coder and uses it to sign commits with Git.
+It requires `jq` to be installed inside your workspace.
+
+This is not recommended if your workspace can be accessed by other/unwanted people, in the case an administrator account on your Coder account gets breached, the attacker could gain access to your workspace and sign commits on your behalf (since the key is stored in the worksace).
+If your Coder account gets breached, the SSH key could also be used on your behalf.
+
+```hcl
+module "git-commit-signing" {
+    source = "https://registry.coder.com/modules/git-commit-signing"
+    agent_id = coder_agent.example.id
+}
+```
--- a/git-commit-signing/README.md
+++ b/git-commit-signing/README.md
@@ -1,7 +1,7 @@
 ---
 display_name: Git commit signing
 description: Configures Git to sign commits using your Coder SSH key
-icon:  ../.icons/git.svg
+icon: ../.icons/git.svg
 maintainer_github: phorcys420
 verified: false
 tags: [helper, git]
--- a/git-commit-signing/README.md
+++ b/git-commit-signing/README.md
@@ -10,7 +10,7 @@ tags: [helper, git]
 # git-commit-signing
 
 This module downloads your SSH key from Coder and uses it to sign commits with Git.
-It requires `jq` to be installed inside your workspace.
+It requires `curl` and `jq` to be installed inside your workspace.
 
 Please observe that using the SSH key that's part of your Coder account for commit signing, means that in the event of a breach of your Coder account, or a malicious admin, someone could perform commit signing pretending to be you.
 
--- a/git-commit-signing/README.md
+++ b/git-commit-signing/README.md
@@ -14,6 +14,8 @@ It requires `curl` and `jq` to be installed inside your workspace.
 
 Please observe that using the SSH key that's part of your Coder account for commit signing, means that in the event of a breach of your Coder account, or a malicious admin, someone could perform commit signing pretending to be you.
 
+This module has a chance of conflicting with the user's dotfiles / the personalize module if one of those has configuration directives that overwrite this module's / each other's git configuration.
+
 ```hcl
 module "git-commit-signing" {
     source = "https://registry.coder.com/modules/git-commit-signing"
--- a/git-commit-signing/README.md
+++ b/git-commit-signing/README.md
@@ -12,7 +12,7 @@ tags: [helper, git]
 This module downloads your SSH key from Coder and uses it to sign commits with Git.
 It requires `jq` to be installed inside your workspace.
 
-This is not recommended if your workspace can be accessed by other/unwanted people, in the case an administrator account on your Coder account gets breached, the attacker could gain access to your workspace and sign commits on your behalf (since the key is stored in the worksace).
+This is not recommended if your workspace can be accessed by other/unwanted people, in the case an administrator account on your Coder account gets breached, the attacker could gain access to your workspace and sign commits on your behalf (since the key is stored in the workspace).
 If your Coder account gets breached, the SSH key could also be used on your behalf.
 
 ```hcl
--- a/git-commit-signing/README.md
+++ b/git-commit-signing/README.md
@@ -12,8 +12,7 @@ tags: [helper, git]
 This module downloads your SSH key from Coder and uses it to sign commits with Git.
 It requires `jq` to be installed inside your workspace.
 
-This is not recommended if your workspace can be accessed by other/unwanted people, in the case an administrator account on your Coder account gets breached, the attacker could gain access to your workspace and sign commits on your behalf (since the key is stored in the workspace).
-If your Coder account gets breached, the SSH key could also be used on your behalf.
+Please observe that using the SSH key that's part of your Coder account for commit signing, means that in the event of a breach of your Coder account, or a malicious admin, someone could perform commit signing pretending to be you.
 
 ```hcl
 module "git-commit-signing" {
--- a/git-commit-signing/main.tf
+++ b/git-commit-signing/main.tf
@@ -9,10 +9,6 @@ terraform {
   }
 }
 
-locals {
-  icon_url = "https://raw.githubusercontent.com/coder/modules/main/.icons/git.svg"
-}
-
 variable "agent_id" {
   type        = string
   description = "The ID of a Coder agent."
@@ -20,7 +16,7 @@ variable "agent_id" {
 
 resource "coder_script" "git-commit-signing" {
   display_name = "Git commit signing"
-  icon         = local.icon_url
+  icon         = "https://raw.githubusercontent.com/coder/modules/main/.icons/git.svg"
 
   script       = file("${path.module}/run.sh")
   run_on_start = true
--- a/git-commit-signing/main.tf
+++ b/git-commit-signing/main.tf
@@ -1,0 +1,29 @@
+terraform {
+  required_version = ">= 1.0"
+
+  required_providers {
+    coder = {
+      source  = "coder/coder"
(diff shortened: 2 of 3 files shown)


git-commit-signing/main.tf
```diff
@@ -0,0 +1,29 @@
...
+    coder = {
+      source  = "coder/coder"
+      version = ">= 0.12"
+    }
+  }
+}
+
+locals {
+  icon_url = "https://raw.githubusercontent.com/coder/modules/main/.icons/git.svg"
+}
```
matifali, line 14:
```suggestion
```
remove this as you only need icon at 1 place.

git-commit-signing/main.tf
```diff
@@ -0,0 +1,29 @@
...
+
+variable "agent_id" {
+  type        = string
+  description = "The ID of a Coder agent."
+}
+
+resource "coder_script" "git-commit-signing" {
+  display_name = "Git commit signing"
+  icon         = local.icon_url
```
matifali, line 23:
```suggestion
  icon         = "/icon/git.svg"
```

git-commit-signing/README.md
```diff
@@ -0,0 +1,65 @@
...
+---
+
+# git-commit-signing
+
+<!-- Describes what this module does -->
+
+```hcl
+module "git-commit-signing" {
+    source = "https://registry.coder.com/modules/git-commit-signing"
+}
```
matifali, line 17:
```suggestion
module "git-commit-signing" {
    source   = "https://registry.coder.com/modules/git-commit-signing"
    agent_id = "coder_agent.example.id" 
}
```
`agent_id` is a required variable for modules using `coder_script`

git-commit-signing/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Run code-server in the background, don't fetch it from GitHub:
+
+```hcl
+module "git-commit-signing" {
+    source = "https://registry.coder.com/modules/git-commit-signing"
+    agent_id = coder_agent.example.id
+    offline = true
+}
+```
```
matifali, line 65:
```suggestion
```
Examples are unrelated

matifali (changes requested):

git-commit-signing/README.md
phorcys420: forgot to commit it

git-commit-signing/README.md
phorcys420: forgot to push

git-commit-signing/README.md
```diff
@@ -0,0 +1,23 @@
...
+tags: [helper, git]
+---
+
+# git-commit-signing
+
+This module downloads your SSH key from Coder and uses it to sign commits with Git.
+It requires `jq` to be installed inside your workspace.
+
+This is not recommended if your workspace can be accessed by other/unwanted people, in the case an administrator account on your Coder account gets breached, the attacker could gain access to your workspace and sign commits on your behalf (since the key is stored in the worksace).
+If your Coder account gets breached, the SSH key could also be used on your behalf.
```
matifali, line 16:
```suggestion
This is not recommended if your workspace can be accessed by other/unwanted people, in the case an administrator account on your Coder account gets breached, the attacker could gain access to your workspace and sign commits on your behalf (since the key is stored in the workspace).
If your Coder account gets breached, the SSH key could also be used on your behalf.
```

matifali (approved):
All good. Can you add tests for testing git and jq? You can also, check at other git based modules tests.

phorcys420:
> All good. Can you add tests for testing git and jq?

I did in [ebf0bf1](https://github.com/coder/modules/pull/94/commits/ebf0bf1b54a1b99ab37925e54d0c7b2ee21da773)

> You can also, check at other git based modules tests.

I did check but as I said there's not really any testing possible because i'd need a valid agent token to ensure it all works out.
and if I wanted to check if ssh key is being added properly instead i would need to add more complexity to the script to allow passing of an arbitrary SSH key.

matifali (dismissed):

matifali:
Thank you, @phorcys420, for the contribution.

git-commit-signing/README.md
```diff
@@ -0,0 +1,23 @@
...
+tags: [helper, git]
+---
+
+# git-commit-signing
+
+This module downloads your SSH key from Coder and uses it to sign commits with Git.
+It requires `jq` to be installed inside your workspace.
+
+This is not recommended if your workspace can be accessed by other/unwanted people, in the case an administrator account on your Coder account gets breached, the attacker could gain access to your workspace and sign commits on your behalf (since the key is stored in the workspace).
+If your Coder account gets breached, the SSH key could also be used on your behalf.
```
mafredri, line 16: I think we could simplify this disclaimer somewhat as I don't think this module actually increases risk by much. The only difference to not using this module is that the SSH key is stored on disk, i.e. your home volume.

If an admin gained access to your workspace, they could still issue the `curl` command used here to get your key. Likewise, if your Coder account is breached, the key can simply be pulled from the server.

git-commit-signing/README.md
```diff
@@ -0,0 +1,23 @@
...
+icon: ../.icons/git.svg
+maintainer_github: phorcys420
+verified: false
+tags: [helper, git]
+---
+
+# git-commit-signing
+
+This module downloads your SSH key from Coder and uses it to sign commits with Git.
+It requires `jq` to be installed inside your workspace.
```
mafredri, line 13: Could mention that it requires `curl` as well. Sometimes only one of `curl` or `wget` is available.

git-commit-signing/README.md
phorcys420: i see what you mean but what i meant is the risk of adding that ssh key as a signing key means that you kinda compromise signing

git-commit-signing/run.sh
```diff
@@ -0,0 +1,36 @@
...
+EOF
+
+chmod -R 400 ~/.ssh/coder
+chmod -R 400 ~/.ssh/coder.pub
+
+echo "Configuring git to use the SSH key"
+
+git config --global gpg.format ssh
+git config --global commit.gpgsign true
+git config --global user.signingkey ~/.ssh/coder
```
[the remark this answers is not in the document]
phorcys420: I guess we should signal it in the README maybe, but isn't that the case with the `git-config` module too ?

matifali (changes requested):

git-commit-signing/run.sh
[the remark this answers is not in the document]
mafredri: Yes, good observation. It's absolutely (going to be) a problem in other modules as well, and `git-config` exhibits exactly the same problem as I described above.

git-commit-signing/README.md
mafredri: Ah, I see what you mean. Perhaps changing the message slightly would make sense then?

> Please observe that using the SSH key that's part of your Coder account for commit signing, means that in the event of a breach of your Coder account, or malicious admin, they could perform commit signing pretending to be you.

Apologies if the wording is bad, I mainly wanted to convey my thought, feel free to use it, a part of it, or not at all. 😄

git-commit-signing/run.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo "Downloading SSH key"
+
+ssh_key=$(curl --request GET \
+        --url "${CODER_AGENT_URL}api/v2/workspaceagents/me/gitsshkey" \
+        --header "Coder-Session-Token: ${CODER_AGENT_TOKEN}")
+
+jq --raw-output ".public_key" > ~/.ssh/coder.pub <<EOF
+$ssh_key
+EOF
```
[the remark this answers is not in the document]
mafredri: I agree, this is a good solution. 👍🏻 Fewer dependencies means more portability.

You could also remove the `sed` dependency via (changes `\n` for literal newline, could use empty string instead):

```
awk -F ':"' '{gsub(/\\n/,"\n"); print $2}'
```

Both `awk` and `sed` are very portable though.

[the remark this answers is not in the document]
phorcys420: I really don't like parsing JSON or other data formats using text parsing tools.
feels like using a regex for XML, it works but it's pretty dirty, I'm really not sure about this one.

git-commit-signing/README.md
phorcys420: No, I actually thing the wording is pretty good and not too long to read like mine.

phorcys420: 3e9e6aad0b0ad9c2569419f3a29ed541cfcf0506

phorcys420: 3e9e6aad0b0ad9c2569419f3a29ed541cfcf0506

git-commit-signing/README.md
phorcys420: bf7f9d04163e9c085a7c57306e4198aa84019fb0

git-commit-signing/README.md
phorcys420: 326fafad027dc4462292a5d3ec84a869d66aa9a3

phorcys420: bf7f9d04163e9c085a7c57306e4198aa84019fb0

git-commit-signing/run.sh
```diff
@@ -0,0 +1,36 @@
...
+
+mkdir -p ~/.ssh
+
+echo "Downloading SSH key"
+
+ssh_key=$(curl --request GET \
+        --url "${CODER_AGENT_URL}api/v2/workspaceagents/me/gitsshkey" \
+        --header "Coder-Session-Token: ${CODER_AGENT_TOKEN}")
+
+jq --raw-output ".public_key" > ~/.ssh/coder.pub <<EOF
```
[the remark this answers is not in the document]
phorcys420: 396078a6f1f3707dc65c8efc395d51a697682b6d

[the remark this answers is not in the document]
phorcys420: 396078a6f1f3707dc65c8efc395d51a697682b6d

git-commit-signing/run.sh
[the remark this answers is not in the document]
phorcys420: 6d747f3416aa2dcf0241b1ec9979405710b55f0d

---

@mafredri what do you think ? I will also create a separate PR for `git-config` once you approve the wording.

phorcys420:
I just changed the directory to `~/.ssh/git-commit-signing` but wouldn't we want it to be something more generic like `~/.ssh/coder-ssh` or something like this if other modules also use the key ?

I was thinking doing that would allow to reuse the key if it already exists but then it might be outdated, we could check for whether the date is prior to boot time or not.
This could also cause conflict between scripts, maybe we should avoid it.

git-commit-signing/run.sh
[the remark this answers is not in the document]
mafredri: I think it looks good. I'd suggest a few more changes so that it flows better when reading:

> This module may conflict with the users dotfiles, or modules like the personalize-module. A conflict may happen if they replace the git config file or have configuration directives that overwrite the ones in this modules.

matifali (approved):

matifali:
Thank you @phorcys420. This is a great contribution.

PR #96, opened 2023-11-06
chore(jetbrans-gateway): update JetBrains IDEs

matifali:
1. Updated IDEs to the latest available version
2. Remove the Community editions as they are not supported by the gateway.

### DIFF
--- a/jetbrains-gateway/main.tf
+++ b/jetbrains-gateway/main.tf
@@ -36,10 +36,10 @@ variable "jetbrains_ides" {
   validation {
     condition = (
       alltrue([
-        for code in var.jetbrains_ides : contains(["IU", "IC", "PS", "WS", "PY", "PC", "CL", "GO", "DB", "RD", "RM"], code)
+        for code in var.jetbrains_ides : contains(["IU", "PS", "WS", "PY", "CL", "GO", "DB", "RD", "RM"], code)
       ])
     )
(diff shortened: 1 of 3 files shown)


jetbrains-gateway/main.tf
```diff
@@ -58,37 +58,27 @@ locals {
...
     },
     "PY" = {
       icon  = "/icon/pycharm.svg",
       name  = "PyCharm Professional",
-      value = jsonencode(["PY", "232.9559.58", "https://download.jetbrains.com/python/pycharm-professional-2023.2.1.tar.gz"])
-    },
-    "PC" = {
-      icon  = "/icon/pycharm.svg",
-      name  = "PyCharm Community",
-      value = jsonencode(["PC", "232.9559.58", "https://download.jetbrains.com/python/pycharm-community-2023.2.1.tar.gz"])
```
stirby, line 86: If we're removing support for PyCharm Community entirely we should update the README to reflect that.

stirby (changes requested):
One comment on updating the README, looks good!

jetbrains-gateway/main.tf
matifali: It was added as a mistake and does not actually work with the gateway. I will update README.md

stirby (approved):
👍

PR #97, opened 2023-11-12
Add JFrog (OAuth) integration module

### LINKED ISSUE #77: Add OAuth to Coder+JFrog module
This will remove the requirement for an admin-level access token in Artifactory. We can use https://github.com/coder/coder/pull/9996 + JFrog's OAuth "integrations"

matifali:
1. Adds a new JFrog OAuth module that supports integration using OAuth flow instead of an admin-level access token and Artifactory terraform provider.
2. Renames the existing `jfrog` module to `jfrog-token`.

closes #77


.images/jfrog-oauth.png
bpmct, line 1: Can you change the template from `Jfrog` to `JFrog`?

Since it an official template and we are partners with them, we should be consistent :)

matifali: The name?

jfrog-oauth/main.tf
```diff
@@ -0,0 +1,77 @@
...
+
+variable "username_field" {
+  type        = string
+  description = "The field to use for the artifactory username. i.e. Coder username or email."
+  default     = "username"
+  validation {
+    condition     = can(regex("^(email|username)$", var.username_field))
+    error_message = "username_field must be either 'email' or 'username'"
+  }
+}
```
bpmct, line 25: Why is this necessary? Can we read the username from the token if its a JWT? [example](https://github.com/coder/customers/issues/248#issuecomment-1679258824)

Perhaps Coder should expose the userinfo/token as a data source

jfrog-token/main.tf
```diff
@@ -52,6 +51,25 @@ For example:
...
+  access_token = var.artifactory_access_token
+}
+
+resource "artifactory_scoped_token" "me" {
+  # This is hacky, but on terraform plan the data source gives empty strings,
+  # which fails validation.
+  username    = length(local.username) > 0 ? local.username : "dummy"
+  scopes      = ["applied-permissions/user"]
+  refreshable = true
+}
```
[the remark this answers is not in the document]
matifali: What do you mean? This module does not use the OAuth provider and makes use of artifactory terrafform provider to get a scoped token.

jfrog-oauth/main.tf
matifali: In our SaaS instance we have usernames set to emails but in the self hosted when we sign up using the GitHub or Google OAuth the usernames are equal to usernames. 

Our `coder_external_auth` data doesn't expose the username and only exposes `access_token`.

jfrog-token/main.tf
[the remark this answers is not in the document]
matifali: Yes I renamed it to remove the confusion.

jfrog-oauth/main.tf
bpmct: If you put the token into http://jwt.io, does it include the user's email?

jfrog-oauth/main.test.ts
```diff
@@ -0,0 +1,17 @@
...
+} from "../test";
+
+describe("jfrog-oauth", async () => {
+  await runTerraformInit(import.meta.dir);
+
+  testRequiredVariables(import.meta.dir, {
+    agent_id: "some-agent-id",
+    jfrog_url: "http://localhost:8081",
+    package_managers: "{}",
+  });
```
bpmct, line 16: non-blocking. perhaps you and @stirby could pair on writing some tests?

bpmct (approved):

jfrog-oauth/main.test.ts
matifali: I am not yet comfortable with these tests as they are more cosmetic. We should probably test E2E with an actual deployment.

PR #98, opened 2023-11-14
Revert "chore(jetbrans-gateway): update JetBrains IDEs"

spikecurtis:
Reverts coder/modules#96

Breaks dev.coder.com like 

```
Error: Error in function call
on .terraform/modules/jetbrains_gateway/main.tf line 129, in resource "coder_app" "gateway":
  129:   display_name = data.coder_parameter.jetbrains_ide.option[index(data.coder_parameter.jetbrains_ide.option.*.value, data.coder_parameter.jetbrains_ide.value)].name
    ├────────────────
    │ data.coder_parameter.jetbrains_ide.option is list of object with 2 elements
    │ data.coder_parameter.jetbrains_ide.value is "[\"GO\",\"232.9921.53\",\"https://download.jetbrains.com/go/goland-2023.2.2.tar.gz\"]"
Call to function "index" failed: item not found.
```

### DIFF
--- a/jetbrains-gateway/README.md
+++ b/jetbrains-gateway/README.md
@@ -17,8 +17,7 @@ module "jetbrains_gateway" {
   agent_id       = coder_agent.example.id
   agent_name     = "example"
   folder         = "/home/coder/example"
-  jetbrains_ides = ["GO", "WS", "IU", "PY", "PS", "CL", "RM", "DB", "RD"]
-  default        = "PY"
+  jetbrains_ides = ["GO", "WS", "IU", "IC", "PY", "PC", "PS", "CL", "RM", "DB", "RD"]
 }
 ```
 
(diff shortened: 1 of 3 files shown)


matifali:
@spikecurtis I am investigating why this breaks down the template as the changes are totally unrelated to the error we get. 
This is again a strong point in favor of that we should have versioning

matifali (approved):
I am approving this to unblock everyone but will soon be merging the required fix.

spikecurtis:
I think what might have happened is that the registry was updated with new options, but the dogfood template still refers to the old versions in it's set of options.

matifali:
@spikecurtis If you see the change, we only use `WS` and `GO` on dogfood, but I removed `PY` and `IC`. So, it should not be related, apparently.

Issue #70, opened 2023-10-02
create a versioning system


matifali:
As discussed in #55, we should have a versioning system to control breaking changes and new behavior of modules.

matifali:
This has been explained in detail in https://github.com/coder/registry.coder.com/issues/7.

PR #99, opened 2023-11-14
chore(jetbrans-gateway): update JetBrains IDEs and remove community editions

matifali:
Another try at #96. I cannot reproduce the error on deployment and dev.coder.com.

It also removes Community editions, `Rider`, and `DataGrip` as they are not supported for Remote Development.

### DIFF
--- a/jetbrains-gateway/main.tf
+++ b/jetbrains-gateway/main.tf
@@ -14,11 +14,6 @@ variable "agent_id" {
   description = "The ID of a Coder agent."
 }
 
-variable "agent_name" {
-  type        = string
-  description = "The name of a Coder agent."
-}
-
 variable "folder" {
   type        = string
   description = "The directory to open in the IDE. e.g. /home/coder/project"
@@ -133,8 +128,8 @@ resource "coder_app" "gateway" {
   url = join("", [
     "jetbrains-gateway://connect#type=coder&workspace=",
     data.coder_workspace.me.name,
-    "&agent=",
-    var.agent_name,
+    "&agent_id=",
+    var.agent_id,
     "&folder=",
     var.folder,
     "&url=",
--- a/jetbrains-gateway/main.tf
+++ b/jetbrains-gateway/main.tf
@@ -36,10 +36,10 @@ variable "jetbrains_ides" {
   validation {
     condition = (
       alltrue([
-        for code in var.jetbrains_ides : contains(["IU", "IC", "PS", "WS", "PY", "PC", "CL", "GO", "DB", "RD", "RM"], code)
+        for code in var.jetbrains_ides : contains(["IU", "PS", "WS", "PY", "CL", "GO", "DB", "RD", "RM"], code)
       ])
     )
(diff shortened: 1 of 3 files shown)


matifali:
@spikecurtis I tested this on my deployment. And it is working fine.

jetbrains-gateway/main.tf
```diff
@@ -58,37 +58,27 @@ locals {
     "GO" = {
       icon  = "/icon/goland.svg",
       name  = "GoLand",
-      value = jsonencode(["GO", "232.9921.53", "https://download.jetbrains.com/go/goland-2023.2.2.tar.gz"])
+      value = jsonencode(["GO", "232.10203.20", "https://download.jetbrains.com/go/goland-2023.2.4.tar.gz"])
```
spikecurtis, line 61: I don't think we can just redefine these options.  Existing workspaces that were built with the old options have the old values stored as parameters.  Then, when trying to rebuild the workspace, it tries to look up the display name and icon using the old value and `index` and this is what is failing.

spikecurtis (dismissed):

jetbrains-gateway/main.tf
matifali: But should not this download the newer module and use the new values? I am confused about why it would use the old values.

spikecurtis: Yeah, I think that without a new template version, Coder doesn't spot that the parameter value from the old build is potentially outdated, so it doesn't trigger reselecting a valid parameter value.

matifali: Could you approve it for now? We can also force update the version

bpmct:
Perhaps a long-term solution is to support versions via template variables? I know versions would also help too

jetbrains-gateway/main.tf
bpmct: > We can also force update the version

So this will break other deployments that use our module?

matifali: yes it will.

matifali:
@bpmct, can you elaborate on how it will work? Modules already use terraform variables to get their inputs.

jetbrains-gateway/main.tf
spikecurtis: How urgent is this update?  

The breakage is pretty confusing, and if your workspace is running you can't update because that runs a stop, which is broken.

I really think we should wait until we have a solution that doesn't break us or others.

matifali: It is not urgent and we can wait for a fix.

matifali:
Blocked until we have #70, as it breaks existing templates using the module.
More details at: https://github.com/coder/registry.coder.com/issues/7

bpmct (approved):

matifali:
### Merge activity

* **Dec 8, 2:34 AM**: @@matifali started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/modules/99).

OLCadlywydd:
I don't know when support was added, but it seems remote development for `Rider` is supported now.

PR #101, opened 2023-11-15
feat: add JFrog access token output to module output

matifali:
This pull request adds a new output to the JFrog modules that allows the template author to access the JFrog access token. This output is named `access_token` and its value is retrieved from the `data.coder_external_auth.jfrog.access_token` and `artifactory_scoped_token.me.access_token` data sources. This change will make it easier for users to integrate their JFrog credentials with other terraform resources.


bpmct (approved):

matifali:
Updated with example on how to use the token from the module.

PR #105, opened 2023-11-26
Add Hashicorp Vault Integration (GitHub)

matifali:
This pull request adds an authentication module with Hashicorp Vault using GitHub external auth. The module allows users to fetch secrets from Vault using the Vault CLI or the Vault API.


vault-github/README.md
```diff
@@ -0,0 +1,74 @@
...
+description: Authenticates with Vault using GitHub
+icon: ../.icons/vault.svg
+maintainer_github: coder
+verified: true
+tags: [helper, integration, vault, github]
+---
+
+# Hashicorp Vault Integration (GitHub)
+
+This module lets you authenticate with [Hashicorp Vault](https://www.vaultproject.io/) in your Coder workspaces using GitHub [external auth](https://coder.com/docs/v2/latest/admin/external-auth).
```
mafredri, line 12:
```suggestion
This module lets you authenticate with [Hashicorp Vault](https://www.vaultproject.io/) in your Coder workspaces using [external auth](https://coder.com/docs/v2/latest/admin/external-auth) for GitHub.
```

It's coders external auth, not GitHubs, so IMO this way it's a bit clearer.

vault-github/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  agent_id     = var.agent_id
+  display_name = "vault"
+  icon         = "/icon/vault.svg"
+  script = templatefile("${path.module}/run.sh", {
+    VAULT_ADDR : var.vault_addr,
+    AUTH_PATH : var.vault_github_auth_path,
+    GITHUB_EXTERNAL_AUTH_ID : data.coder_external_auth.github.id,
+    VERSION : var.vault_cli_version,
+  })
+  run_on_start = true
```
mafredri, line 55: Should we add [`start_blocks_login`](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/script#start_blocks_login) so that it's guaranteed to be set up before use?

vault-github/run.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
+
+BOLD='\033[0;1m'
+VAULT_ADDR=${VAULT_ADDR}
+VERSION=${VERSION}
+AUTH_PATH=${AUTH_PATH}
+GITHUB_EXTERNAL_AUTH_ID=${GITHUB_EXTERNAL_AUTH_ID}
+
+# Check if vault is installed
+if ! command -v vault &>/dev/null; then
```
mafredri, line 10: Should we verify it's the correct program and version and install a new one if not?

vault-github/run.sh
```diff
@@ -0,0 +1,79 @@
...
+    if [ "${VERSION}" = "latest" ]; then
+        INSTALL_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -oP '[0-9]+\.[0-9]+\.[0-9]' | tr -d '<>' | head -1)
+    else
+        INSTALL_VERSION=$VERSION
+    fi
+
+    # download vault
+    wget -q -O vault.zip https://releases.hashicorp.com/vault/$${INSTALL_VERSION}/vault_$${INSTALL_VERSION}_linux_amd64.zip
+    unzip vault.zip
+    sudo mv vault /usr/local/bin
```
mafredri, line 32: Perhaps `sudo` isn't guaranteed? How about installing into `~/.local/bin` or `~/.local/share/vault[/bin]` instead and updating PATH?

```sh
export PATH="${HOME}/.local/share/vault/bin:${PATH}"
```

vault-github/run.sh
```diff
@@ -0,0 +1,79 @@
...
+    # check if wget is installed
+    if ! command -v wget &>/dev/null; then
+        printf "wget is not installed. Please install wget in your image.\n"
+        exit 1
+    fi
+    # check if unzip is installed
+    if ! command -v unzip &>/dev/null; then
+        printf "unzip is not installed. Please install unzip in your image.\n"
+        exit 1
+    fi
```
mafredri, line 21: It would be nice to allow `curl` and `busybox unzip` as fallbacks if either wget/unzip aren't available. Perhaps a matter of future improvement. We could consider baking it into the agent as well (hidden command, `coder tools wget`, `coder tools unzip`, etc.)

vault-github/run.sh
```diff
@@ -0,0 +1,79 @@
...
+vault status
+
+# Login to Vault to using GitHub token
+printf "🔑 Logging in to Vault ...\n\n"
+vault login -no-print -method=github -path=/$AUTH_PATH token=$GITHUB_TOKEN
+
+# Add VAULT_ADDR to shell login scripts if not already present e.g. .bashrc, .zshrc
+# This is a temporary fix and will be replaced with https://github.com/coder/coder/issues/10166
+# bash
+if [[ -f ~/.bashrc ]] && ! grep -q "VAULT_ADDR" ~/.bashrc; then
```
mafredri, line 61: These files may be initially missing, but set up by user dotfiles later. I don't have a good alternative solution, but there's a good chance this will only work on the second start or something similar. If we create the files when missing, the users dotfiles may overwrite them, so same issue (works after second start).

I can't really think of a fix other than targeting `/etc` files for the shells. But that also requires `sudo` and might have other unforeseen issues.

vault-github/run.sh
```diff
@@ -0,0 +1,79 @@
...
+if [[ -f ~/.zshrc ]] && ! grep -q "VAULT_ADDR" ~/.zshrc; then
+    printf "\nAdding VAULT_ADDR to ~/.zshrc ...\n"
+    echo "export VAULT_ADDR=$VAULT_ADDR" >>~/.zshrc
+fi
+
+# fish
+if [[ -f ~/.config/fish/config.fish ]] && ! grep -q "VAULT_ADDR" ~/.config/fish/config.fish; then
+    printf "\nAdding VAULT_ADDR to ~/.config/fish/config.fish ...\n"
+    echo "set -x VAULT_ADDR $VAULT_ADDR" >>~/.config/fish/config.fish
+fi
```
mafredri, line 76: If none of the files are present.. print a big warning?

vault-github/README.md
```diff
@@ -0,0 +1,74 @@
...
+Then you can use the Vault CLI in your workspaces to fetch secrets from Vault:
+
+```shell
+vault kv get -mount=secret my-secret
+```
+
+or using the Vault API:
+
+```shell
+curl -H "X-Vault-Token: $VAULT_TOKEN" -X GET $VAULT_ADDR/v1/secret/data/my-secret
```
mafredri, line 31: More quoting, just to be safe.. 😄

```suggestion
curl -H "X-Vault-Token: ${VAULT_TOKEN}" -X GET "${VAULT_ADDR}/v1/secret/data/my-secret"
```

vault-github/run.sh
```diff
@@ -0,0 +1,79 @@
...
+        exit 1
+    fi
+    # check if unzip is installed
+    if ! command -v unzip &>/dev/null; then
+        printf "unzip is not installed. Please install unzip in your image.\n"
+        exit 1
+    fi
+    # check if VERSION is latest
+    if [ "${VERSION}" = "latest" ]; then
+        INSTALL_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -oP '[0-9]+\.[0-9]+\.[0-9]' | tr -d '<>' | head -1)
```
mafredri, line 24: Mix of both `curl` and `wget`, I'd recommend to prefer one or support both.

mafredri (commented):
Looks pretty good to me. Readme could still list how to make this work when automatic (e.g. add to *rc files) install fails. But considering it's a temporary workaround, perhaps not that important.

vault-github/main.tf
matifali: yes will do

vault-github/run.sh
matifali: We are already checking for the binary in the path but not the version. I can add a check for the version, too.

vault-github/run.sh
matifali: We do not have an easy way to manipulate the path yet. This would have to be managed using a shell init files too :/

vault-github/run.sh
matifali: How can I fallback to the busybox unzip?

vault-github/run.sh
matifali: Yes unfortunatly. Hopefully we get rid of them when we have `coder_env`

bpmct:
Atif, this is awesome. I'm a bit nervous to merge things that modify bashrc to set vars. Perhaps @mafredri can help us find another abstraction to make this more solid?

vault-github/run.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
mafredri, line 1: This improves portability, but there's some bashisms we should get rid off. PS. Does this repo already use shftm/shellcheck? If not, might be a good addition.

```suggestion
#!/usr/bin/env sh
```

vault-github/run.sh
```diff
@@ -0,0 +1,85 @@
...
+    # Check if unzip or busybox is installed
+    if ! command -v unzip &>/dev/null; then
+        if command -v busybox &>/dev/null && busybox --list | grep -q '^unzip$'; then
+            alias unzip='busybox unzip'
+            printf "Using busybox unzip.\n"
+        else
+            printf "unzip is not installed and busybox unzip is not available. Please install unzip in your image.\n"
+            exit 1
+        fi
+    fi
```
mafredri, line 44: I think the alias won't work here, and pretty sure not in POSIX shells (sh).

How about defining the following functions instead?

```sh
fetch() {
    dest="$1"
    url="$2"
    if command -v curl; then
        command curl -sSL --fail "${url}" -o "${dest}"
    elif command -v wget; then
        command wget -O "${dest}" "${url}"
    elif command -v busybox; then
        busybox wget -O "${dest}" "${url}"
    else
        printf "curl, wget, or busybox is not installed. Please install curl or wget in your image.\n"
        exit 1
    fi
}

unzip() {
    if command -v unzip; then
        command unzip "$@"
    elif command -v busybox; then
        busybox unzip "$@"
    else
        printf "unzip or busybox is not installed. Please install unzip in your image.\n"
        exit 1
    fi
}
```

Then it can be used as `fetch vault.zip "https://..."` and `unzip -o vault.zip` and keep the "main code" clean. It'll also take care of any alias problems.

vault-github/run.sh
```diff
@@ -0,0 +1,85 @@
...
+    VERSION=$LATEST_VERSION
+fi
+
+# Check if vault is installed and has the correct version
+installation_needed=1
+if command -v vault &>/dev/null; then
+    CURRENT_VERSION=$(vault version | grep -oE '[0-9]+\.[0-9]+\.[0-9]+')
+    if [ "$CURRENT_VERSION" = "$VERSION" ]; then
+        printf "Vault version $CURRENT_VERSION is already installed and up-to-date.\n\n"
```
mafredri, line 29: Variables with `printf` is not recommended:

```suggestion
        printf "Vault version %s is already installed and up-to-date.\n\n" "$CURRENT_VERSION"
```

mafredri:
> Atif, this is awesome. I'm a bit nervous to merge things that modify bashrc to set vars. Perhaps @mafredri can help us find another abstraction to make this more solid?

The only option I can think of is to wrap the `vault` binary. Consider the following script in `/usr/local/bin/vault`.

```sh
#!/usr/bin/env sh

export VAULT_ADDR=${VAULT_ADDR:-https://my-addr}
exec /path/to/real/binary/thats/not/in/PATH/vault
```

It's not ideal either, though, since it won't expose the address. And if vault is already present, we need to move it or place the wrapper in a higher priority path.

Until we ship `coder_env`, I'm afraid I don't have a better alternative. The only other way to get a modification in there currently is to either modify user (as now) or system shell init files.

matifali:
@mafredri I am removing this logic of modifying shell files for now and suggesting the user uses an `env` in `coder_agent` block.
We can remove this later on.

vault-github/run.sh
```diff
@@ -0,0 +1,97 @@
...
+  if sudo mv vault /usr/local/bin/vault 2> /dev/null; then
+    printf "Vault installed successfully!\n\n"
+  else
+    mkdir -p ~/.local/bin
+    mv vault ~/.local/bin/vault
+    if [ ! -f ~/.local/bin/vault ]; then
+      printf "Failed to move Vault to local bin.\n"
+      exit 1
+    fi
+    printf "Please add ~/.local/bin to your PATH to use vault CLI.\n"
```
mafredri, line 79: Nice fallback! 👍🏻

mafredri (approved):
Except for the `set -e` gotcha, looking great 👍🏻

vault-github/run.sh
```diff
@@ -0,0 +1,97 @@
...
+
+if [ $${installation_needed} -eq 1 ]; then
+  # Download and install Vault
+  if [ -z "$${CURRENT_VERSION}" ]; then
+    printf "Installing Vault CLI ...\n\n"
+  else
+    printf "Upgrading Vault CLI from version %s to %s ...\n\n" "$${CURRENT_VERSION}" "$${VERSION}"
+  fi
+  fetch vault.zip "https://releases.hashicorp.com/vault/$${VERSION}/vault_$${VERSION}_linux_amd64.zip"
+  if [ $? -ne 0 ]; then
```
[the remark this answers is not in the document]
matifali: I should remove the `set -e`. Thanks

PR #106, opened 2023-11-28
Add formatting check for shell scripts

matifali:
## Context

This pull request adds a formatting check for shell scripts. It ensures that the shell scripts follow a consistent formatting style. This helps improve code readability and maintainability. The formatting check is performed using the Prettier plugin for shell scripts.

## Possible issues

We must be cautious with these scripts as we populate them with variables from terraform using [`templatefile`](https://developer.hashicorp.com/terraform/language/functions/templatefile) function. This expects a map of vars passed for all variables used as `${VAR}.` To use the variables local to the script, we need to escape them with `$${LOCAL_VAR}`; otherwise, terraform tries to replace them and errors out as it can not find a variable with that name in the map. 

### We also need a method to test all module scripts using automated tests. 

Suggested by @mafredi


mafredri:
A few sh script still seem unformatted (`.sample`, `new.sh`).

> We must be cautious with these scripts as we populate them with variables from terraform using templatefile function.

I would suggest establishing a consistent practice for this. Perhaps we always prefix the inputs from terraform as `TF_`, `TFVAR_`, `_MY_VAR`, etc and begin each script with transferring them to shell vars:

```sh
#!/usr/bin/env

MY_VAR1=${_MY_VAR1}
MY_VAR2=${_MY_VAR2}

echo $${MY_VAR2}
```

Then we can add linting (basically just a script?) that after the initial assignment block, we disallow `${}` to prevent errors and you must either use `$${MY_VAR1}` or `$MY_VAR1`.

Or actually, I'd like to turn that upside-down and have us post-process scripts to turn `${}` into `$${}` automatically. This way we can write scripts normally without taking into account terraform syntax and have shell tools work as expected (e.g. `shellcheck`).

Also, adding `shellcheck` for checking script correctness could still be a big win, in addition to this.

mafredri (approved):

matifali:
@mafredri You mean we run `shellcheck` on original scripts. Where do we post-process?

matifali:
I added `shellcheck` and solved the errors while suppressing `warnings`, `info`, and `style` checks.

matifali:
> Yes, shellcheck on original scripts. However, I'm pretty sure $${} breaks shellcheck, which is why we need either:

yes. `shellcheck complains that these variables are not used.`

> ```shell
> MY_VAR1=${_MY_VAR1}
> MY_VAR2=${_MY_VAR2}
> ```
 for this, I get self-assignment

mafredri:
I imagine we can use the `# shellcheck disable=SCxxx` and `# shellcheck enable=SCxxx` comments around the tf var assignments.

However, if the shellcheck integration is causing too much trouble, we can punt it for later.

matifali:
Removed `shellcheck` for now

PR #107, opened 2023-11-28
Add missing URL to VS Code desktop module

code-asher:
Without this the plugin will only work if the user has happened to log in before and that URL was previously saved (and it is the right URL for whatever workspace they are trying to connect).  Otherwise, they get an error about localhost (or if the URL is for a different deployment they would get some other issues).

Fixes https://github.com/coder/coder/issues/10896.


matifali (approved):
Thank you

PR #115, opened 2023-12-25
feat(jfrog): add JFrog vscode extension, CLI completion and docker support

matifali:
1. Adds JFrog extension configuration and installation.
2. Add support for configuring docker.
3. Adapt `.npmrc` to work with version >=9

- [x] Check for docker command before the docker login

Related https://github.com/coder/coder/issues/11338


jfrog-oauth/run.sh
```diff
@@ -47,12 +53,75 @@ index-url = https://${ARTIFACTORY_USERNAME}:${ARTIFACTORY_ACCESS_TOKEN}@${JFROG_
...
+  echo "📦 Installing JFrog extension..."
+  /tmp/code-server/bin/code-server --install-extension jfrog.jfrog-vscode-extension
+  echo "🥳 JFrog extension installed!"
+else
+  echo "🤔 Skipping JFrog extension installation. Set configure_code_server to true to install the JFrog extension."
+fi
+
+# Configure the JFrog CLI completion
+echo "📦 Configuring JFrog CLI completion..."
+## Get the user's shell
```
mafredri, line 106:
```suggestion
# Get the user's shell
```

jfrog-oauth/run.sh
```diff
@@ -47,12 +53,75 @@ index-url = https://${ARTIFACTORY_USERNAME}:${ARTIFACTORY_ACCESS_TOKEN}@${JFROG_
...
+SHELLNAME=$(grep "^$USER" /etc/passwd | awk -F':' '{print $7}' | awk -F'/' '{print $NF}')
+## Generate the completion script
```
mafredri, line 108:
```suggestion
# Generate the completion script
```

jfrog-oauth/run.sh
```diff
@@ -47,12 +53,75 @@ index-url = https://${ARTIFACTORY_USERNAME}:${ARTIFACTORY_ACCESS_TOKEN}@${JFROG_
...
+jf completion $SHELLNAME --install
+## Add the completion script to the user's shell profile
```
mafredri, line 110:
```suggestion
# Add the completion script to the user's shell profile
```

jfrog-oauth/run.sh
```diff
@@ -47,12 +53,75 @@ index-url = https://${ARTIFACTORY_USERNAME}:${ARTIFACTORY_ACCESS_TOKEN}@${JFROG_
...
+if [ "$SHELLNAME" == "bash" ]; then
+  if ! grep -q "# jf CLI shell completion" ~/.bashrc; then
+    echo "" >> ~/.bashrc
+    echo "# jf CLI shell completion" >> ~/.bashrc
```
mafredri, line 114: Small suggestion to better highlight the block and also mention where it came from.

```sh
# BEGIN: jf CLI shell completion (added by coder module jfrog-oauth)
...
# END: jf CLI shell completion
```

jfrog-oauth/run.sh
```diff
@@ -47,12 +53,75 @@ index-url = https://${ARTIFACTORY_USERNAME}:${ARTIFACTORY_ACCESS_TOKEN}@${JFROG_
...
+    echo "source $HOME/.jfrog/jfrog_bash_completion" >> ~/.bashrc
```
mafredri, line 115:
```suggestion
    echo 'source "$HOME/.jfrog/jfrog_bash_completion"' >> ~/.bashrc
```

Two changes, 1) we leave `$HOME` unresolved, this makes the file more portable and 2) the path _may_ contain spaces (paranoia mode).

(Same below, for zsh.)

jfrog-token/README.md
```diff
@@ -72,6 +72,25 @@ go get github.com/golang/example/hello
...
+
+```hcl
+module "jfrog" {
+  source = "https://registry.coder.com/modules/jfrog-token"
+  agent_id = coder_agent.example.id
+  jfrog_url = "https://XXXX.jfrog.io"
+  artifactory_access_token = var.artifactory_access_token
+  configure_code_server = true
+  package_managers = {
+    "npm": "npm",Add JFrog extension configuration for code-server
```
mafredri, line 87: Please apply all suggestions to this module as well, where appropriate.

jfrog-oauth/run.sh
```diff
@@ -7,38 +7,44 @@ if command -v jf > /dev/null 2>&1; then
...
+  jf npmc --global --repo-resolve "${REPOSITORY_NPM}"
   cat << EOF > ~/.npmrc
-email = ${ARTIFACTORY_EMAIL}
-registry = ${JFROG_URL}/artifactory/api/npm/${REPOSITORY_NPM}
+email=${ARTIFACTORY_EMAIL}
+registry=${JFROG_URL}/artifactory/api/npm/${REPOSITORY_NPM}
 EOF
-  jf rt curl /api/npm/auth >> ~/.npmrc
+  # if npm version is greater than or equal to 9.0.0, use the new .npmrc format
+  if [ "$(npm --version | cut -d. -f1)" -ge 9 ]; then
```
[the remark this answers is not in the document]
matifali: I agree I will update

jfrog-oauth/run.sh
```diff
@@ -47,12 +53,75 @@ index-url = https://${ARTIFACTORY_USERNAME}:${ARTIFACTORY_ACCESS_TOKEN}@${JFROG_
...
+  }
+}
+EOF
+fi
+
+# Install the JFrog vscode extension for code-server.
+if [ "${CONFIGURE_CODE_SERVER}" == "true" ]; then
+  while ! [ -x /tmp/code-server/bin/code-server ]; do
+    counter=0
+    if [ $counter -eq 30 ]; then
```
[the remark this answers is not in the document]
matifali: yes, I agree. Having a mechanism for dependence is probably the best base. For now we can modify the code-server module to create status files. I am not sure how we can check the insatlling status.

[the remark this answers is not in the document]
matifali: Also, should we condition the part of the module or the whole module? I think your idea can be extended to have a common input for all modules `depends_on = module_name` where we check for file `/tmp/.${module_name}-success` before we start runing `coder_script`

jfrog-oauth/run.sh
matifali: I was doing it Markdown style as sub headings of level 2 :D

jfrog-oauth/run.sh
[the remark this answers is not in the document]
matifali: I am going to test _authToken support for older versions of npm, and if it supports the same structure we can probably remove this version check altogether.

jfrog-oauth/run.sh
```diff
@@ -7,38 +7,44 @@ if command -v jf > /dev/null 2>&1; then
...
-registry = ${JFROG_URL}/artifactory/api/npm/${REPOSITORY_NPM}
+email=${ARTIFACTORY_EMAIL}
+registry=${JFROG_URL}/artifactory/api/npm/${REPOSITORY_NPM}
 EOF
-  jf rt curl /api/npm/auth >> ~/.npmrc
+  # if npm version is greater than or equal to 9.0.0, use the new .npmrc format
+  if [ "$(npm --version | cut -d. -f1)" -ge 9 ]; then
+    echo "//${JFROG_HOST}/artifactory/api/npm/${REPOSITORY_NPM}/:_authToken=${ARTIFACTORY_ACCESS_TOKEN}" >> ~/.npmrc
+  else
+    echo "_auth=$(echo -n '${ARTIFACTORY_USERNAME}:${ARTIFACTORY_ACCESS_TOKEN}' | base64)" >> ~/.npmrc
```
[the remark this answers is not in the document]
matifali: Good catch. This is not intentional. It's probably a Copilot thing I looked over. I will fix.

jfrog-oauth/run.sh
matifali: Also, I have noticed that for some docker workspaces, the `.bashrc` and `.zshrc` files do not exist at creation.

jfrog-oauth/run.sh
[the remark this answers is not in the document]
mafredri: If there's a way to reference another modules resource, I think that would work well for this use case. But I'm not sure how terraform would behave in this case. Maybe we need both a module and a script level dependency so that terraform can exit early if only one module is used without its dependency?

Or do we automatically pull in dependencies without them being expplicitly declared?

Anyway, as for how the script part might look, maybe something like this?

```tf
// code-server
resource "coder_script" "install" {}

// my-module
resource "coder_script" "takes_a_while" {}

resource "coder_script" "finalize" {
	depends_on = [code-server.coder_script.install, coder_script.takes_a_while]
}
```

Again, if we can refer like this, it'd be ideal, otherwise we may need to use strings and define the dependency in runtime? Perhaps it would be a good idea to think about what other use-cases we have, and maybe write a proposal? This might also tie a little bit into https://github.com/coder/coder/issues/11131?

jfrog-oauth/README.md
```diff
@@ -106,6 +105,25 @@ jf pip install requests
...
+
+The [JFrog extension](https://open-vsx.org/extension/JFrog/jfrog-vscode-extension) for VS Code allows you to interact with Artifactory from within the IDE.
+
+```hcl
+module "jfrog" {
+  source = "https://registry.coder.com/modules/jfrog-oauth"
+  agent_id = coder_agent.example.id
+  jfrog_url = "https://jfrog.example.com"
+  username_field = "username" # If you are using GitHub to login to both Coder and Artifactory, use username_field = "username"
+  configure_code_server = true # Add JFrog extension configuration for code-server
```
bpmct, line 118:
```suggestion
  configure_code_server = true # Add JFrog extension configuration for code-server
```

thoughts on install_extension = true?

jfrog-oauth/run.sh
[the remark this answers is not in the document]
bpmct: In the meantime, I'd rather not use status files, but instead wait for the process to be started or a successful curl or something.

bpmct (approved):

bpmct:
Will this break old versions of the extension? If so, would like to hold off until we add support for versioning

jfrog-oauth/README.md
bpmct: On second thought, I think `configure_code_server` is better since it is specific to code-server, not other IDEs such as VS Code

jfrog-oauth/run.sh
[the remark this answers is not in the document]
matifali: For now I am going to merge them without any dependence. Anyone?

matifali:
@bpmct I tried to make it so as not to break old versions by not introducing any new required variables or renaming previous variables.

jfrog-oauth/run.sh
matifali: handled this.

PR #132, opened 2024-01-26
feat!(git-clone): change `path` input to `base_dir` and return `repo_dir` as output

matifali:
After #124, we now automatically append the `path` with the `repo` name, so the example should be changed to pass only the base directory as the `path` input.

### DIFF
--- a/git-clone/README.md
+++ b/git-clone/README.md
@@ -38,6 +38,6 @@ module "git-clone" {
   version  = "1.0.0"
   agent_id = coder_agent.example.id
   url      = "https://github.com/coder/coder"
-  path     = "~/projects/coder/coder"
+  path     = "~/projects/coder"
 }
 ```
--- a/git-clone/README.md
+++ b/git-clone/README.md
@@ -9,7 +9,7 @@ tags: [git, helper]
 
 # Git Clone
 
-This module allows you to automatically clone a repository by URL and skip if it exists in the path provided.
+This module allows you to automatically clone a repository by URL and skip if it exists in the base directory provided.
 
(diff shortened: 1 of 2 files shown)


mafredri (approved):
The fix makes sense.

I wonder if `dest` / `destination` would be a more suitable name for the property now, though? It can be read as "we do a git clone and place it in destination" (same as if you were in `dest` when running `git clone` manually). I'm not sure what `path` conveys to me in this context, is it the place we clone, is it the end name of the cloned repo, etc.

I think users would also appreciate being able to override the folder name that's created.

If we draw parallels to how `git` is used:

```
# Both create `modules` in pwd.
git clone https://github.com/coder/modules
git clone https://github.com/coder/modules.git

# Creates `bar` in `foo` in `pwd`.
git clone https://github.com/coder/modules.git foo/bar
```

Git calls the this extra parameter `directory` that renames the folder. I think we could get away with calling it `name`, too.

Just some thoughts.

matifali:
@mafredri Yes, I agree, but that would be a breaking change. We do have versioning now, and Depedanbot should handle version bumps, too. Unfortunately, Dependabot can not suggest a change of input variable names. That is why I am avoiding renaming inputs as this is out top used module.

mafredri:
@matifali I agree about breaking, but to be fair, this change is breaking too. If someone had previously configured a `path`, it would change due to #124m, no?

matifali:
@mafredri Hmm, yes. So let's change it. What about `base_dir`?

mafredri (approved):
Good choice for param, LGTM!

PR #133, opened 2024-01-26
feat(dotfiles): add ability to apply dotfiles as root

phorcys420:
In my template ([basic-env](https://github.com/uwu/basic-env)), dotfiles are also applied to root.
I like doing this because whenever switching to a root shell for more than one command (e.g installing packages, etc), my shell is consistent with my user.

I have added a `root` parameter to the module which will apply the dotfiles using `sudo`.

---

```tf
module "dotfiles" {
  source = "git::https://github.com/phorcys420/coder-modules.git//dotfiles?ref=dotfiles-root"
  agent_id = coder_agent.dev.id
  root = true
}
```

### DIFF
--- a/dotfiles/main.tf
+++ b/dotfiles/main.tf
@@ -25,13 +25,10 @@ data "coder_parameter" "dotfiles_uri" {
 }
 
 resource "coder_script" "personalize" {
-  agent_id     = var.agent_id
-  script       = <<-EOT
-    DOTFILES_URI="${data.coder_parameter.dotfiles_uri.value}"
-    if [ -n "$${DOTFILES_URI// }" ]; then
-    coder dotfiles "$DOTFILES_URI" -y 2>&1 | tee -a ~/.dotfiles.log
-    fi
-    EOT
+  agent_id = var.agent_id
+  script = templatefile("${path.module}/run.sh", {
+    DOTFILES_URI : data.coder_parameter.dotfiles_uri.value,
+  })
   display_name = "Dotfiles"
   icon         = "/icon/dotfiles.svg"
   run_on_start = true
--- a/dotfiles/run.sh
+++ b/dotfiles/run.sh
@@ -1,0 +1,7 @@
+DOTFILES_URI="${DOTFILES_URI}"
+ROOT="${ROOT}"
+
+if [ -n "$${DOTFILES_URI// }" ]; then
+  echo "✨ Applying dotfiles for user $USER"
+  coder dotfiles "$DOTFILES_URI" -y 2>&1 | tee -a ~/.dotfiles.log
+fi


matifali (commented):
Hmm. Can you run `bun fmt` and `bun lint` to fix the formatting issues?

dotfiles/run.sh
```diff
@@ -0,0 +1,12 @@
+DOTFILES_URI="${DOTFILES_URI}"
+ROOT="${ROOT}"
+
+if [ -n "$${DOTFILES_URI// }" ]; then
+    echo "✨ Applying dotfiles for user $USER"
+    coder dotfiles "$DOTFILES_URI" -y 2>&1 | tee -a ~/.dotfiles.log
+
+    if [ "$ROOT" = "true" ]; then
+        echo "✨ Applying dotfiles for user root"
+        sudo $(which coder) dotfiles "$DOTFILES_URI" -y 2>&1 | tee -a ~/.dotfiles-root.log
```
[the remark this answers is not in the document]
phorcys420: I didn't do it because `tee` is getting it's output from sudo, not from the actual coder CLI, which means that `tee` is not running as root and therefore can't write in `/root`.

We could do something like `sudo sh -c "$(which coder) dotfiles \"$DOTFILES_URI\" -y 2>&1 | tee -a /root/.dotfiles.log"` but I don't know if it matters that much, what do you think ?

phorcys420:
> Hmm. Can you run `bun fmt` and `bun lint` to fix the formatting issues?

Yes, sorry, I'm not used to formatting every commit yet.

phorcys420:
it seems like the CI has more formatting/linting rules than my local branch does, [e75b62f](https://github.com/coder/modules/pull/133/commits/e75b62fabc00f0f714432707107be29956a414d9) was all good on my end (`bun fmt` and `bun lint` both exited without modifying anything or erroring yet the CI still fails)

phorcys420:
lol, looks like i messed up, let me do some cleaning up.

phorcys420:
sorry for all the noise, now this should be good.

dotfiles/run.sh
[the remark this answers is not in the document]
matifali: Probably yes. I don't think it's a common use case. So won't affect many people.

dotfiles/README.md
```diff
@@ -11,10 +11,33 @@ tags: [helper]
...
+```
+
+### Apply dotfiles to the current user and root
+
+```tf
+module "dotfiles" {
+  source   = "registry.coder.com/modules/dotfiles/coder"
+  version  = "1.0.0"
+  agent_id = coder_agent.example.id
+  root     = true
```
mafredri, line 41: I feel this param/behavior is a bit weird, I would expect these dotfiles to _only_ to be applied to root.

I have two alternative proposals:
1. Rename this to `user = "root"`, and run with `sudo -u $USER ...` (change behavior to only apply to this one user)
2. Rename this to `extra_users = ["root"]`, this is more flexible and will cover more use-cases, and the "extra" makes it obvious that this will be applied to the standard user _and_ to root.

phorcys420: I think using `user = ?` contrains you to have a module definition per user, which kinda sucks, which is also why I preferred for `root = true` to be extra, but I do agree, it is unclear.
I prefer the `extra_users` way and I have thought about it in the past, but I thought "who's gonna apply to more users than `coder` and `root`?".

I will update the PR to support an `extra_users` array.

mafredri: I don’t think the 1 definition/user constraint is bad per-se, is there a drawback I’m not thinking of?

But I’m fine with either approach.

phorcys420: > I don’t think the 1 definition/user constraint is bad per-se, is there a drawback I’m not thinking of?

Well, I just don't like the idea that the template administrator will have to copy it again, I think that's bad UX.

```tf
module "dotfiles" {
  source = "git::https://github.com/phorcys420/coder-modules.git//dotfiles?ref=dotfiles-root"
  agent_id = coder_agent.dev.id
  extra_users = ["root"]
}
```

is cleaner to me than

```tf
module "dotfiles" {
  source = "git::https://github.com/phorcys420/coder-modules.git//dotfiles?ref=dotfiles-root"
  agent_id = coder_agent.dev.id
}

module "dotfiles-root" {
  source = "git::https://github.com/phorcys420/coder-modules.git//dotfiles?ref=dotfiles-root"
  agent_id = coder_agent.dev.id
  user = "root"
}
```

mafredri: @phorcys420 great insight, thanks for sharing.

I kind of like `user = "root"` because `extra_users` doesn't work with multiple module definitions. But I also don't like forcing two parameters on the user.

Is this possible?

```tf
module "dotfiles" {
  source = "git::https://github.com/phorcys420/coder-modules.git//dotfiles?ref=dotfiles-root"
  agent_id = coder_agent.dev.id
  # exports dotfiles_uri or similar
}

module "dotfiles-root" {
  source = "git::https://github.com/phorcys420/coder-modules.git//dotfiles?ref=dotfiles-root"
  dotfiles_uri = module.dotfiles.dotfiles_uri # when dotfiles_uri is defined, no parameter is created
  agent_id = coder_agent.dev.id
  user = "root"
}
```

Alternatively, the best option for now may be to just rename `root = true` to `also_apply_to_root = true`, or similar. That would convey the behavior better.

phorcys420: I like this way better, with the optional `dotfiles_uri` and `user`, i'll play around with it :-)

dotfiles/run.sh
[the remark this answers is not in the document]
phorcys420: given how this PR is going with custom users rather than arbitrary root, I will make it so that it dynamically resolves the user's home and write to it

phorcys420:
I didn't mean to close, i'm starting off of a clean slate, which is probably what triggered it to close.

dotfiles/README.md
phorcys420: I have added a `dotfiles_uri` optional parameter that, if passed, hides the `coder_parameter` that asks the end user for their dotfiles URI. (873f057)

I will do the same thing for a `user` variable.

phorcys420:
hey @mafredri, while writing this PR, I realized that the script running the dotfiles is named "personalize", do you want me to rename it to "dotfiles" in this PR even though it's not really it's scope, or should I open another PR ? to fix it

```tf
resource "coder_script" "personalize" {
  agent_id = var.agent_id
  script = templatefile("${path.module}/run.sh", {
    DOTFILES_URI : local.dotfiles_uri,
  })
  display_name = "Dotfiles"
  icon         = "/icon/dotfiles.svg"
  run_on_start = true
}
```

dotfiles/main.tf
```diff
@@ -14,7 +14,23 @@ variable "agent_id" {
   description = "The ID of a Coder agent."
 }
 
+variable "dotfiles_uri" {
+  type        = string
+  description = "The URL to a dotfiles repository. (optional)"
+
+  default = null
+}
+
```
matifali, line 23: Why the variable?

phorcys420: see https://github.com/coder/modules/pull/133#discussion_r1470937741

phorcys420:
sorry for all the noise, but given that terraform modules can only be loaded via an URL, I have to push debug commits and then force-push to clean my branch.

here is an example of what you can do with this PR :
```tf
module "dotfiles" {
  source = "git::https://github.com/phorcys420/coder-modules.git//dotfiles?ref=dotfiles-root"
  agent_id = coder_agent.dev.id
  #dotfiles_uri = "<defined by template admin>"
}

module "dotfiles-root" {
  source = "git::https://github.com/phorcys420/coder-modules.git//dotfiles?ref=dotfiles-root"
  agent_id = coder_agent.dev.id
  user = "root"
  dotfiles_uri = module.dotfiles.dotfiles_uri # "<defined by template admin>" if uncommented, or the value the end user provided in template settings
}
```

I have tested all the following scenarios and they are working :
1. single module without `dotfiles_uri` defined
   - asks the user on template creation
   - applies dotfiles to the default user (`coder`)
2. single module with `dotfiles_uri` defined
   - doesn't ask the user on template creation
   - applies dotfiles to the default user (`coder`)
3. single module without `dotfiles_uri` defined with `user = "root"`
   - asks the user on template creation
   - applies dotfiles to the `root` user
4. single module with `dotfiles_uri` defined with `user = "root"
   - doesn't ask the user on template creation
   - applies dotfiles to the `root` user
5. module without `dotfiles_uri` defined + module with `dotfiles_uri = module.dotfiles.dotfiles_uri`
   - asks the user on template creation once and uses the same URI when applying both times
   - applies dotfiles to the default user (`coder`)
6. module with `dotfiles_uri = "<defined by template admin>"` + module with `dotfiles_uri = module.dotfiles.dotfiles_uri`
   - doesn't ask the user at all
   - applies dotfiles to the default user (`coder`)

phorcys420:
@mafredri I would like your input on these things.

---

`main.tf`
```tf
variable "dotfiles_uri" {
  type        = string
  description = "The URL to a dotfiles repository. (optional)"

  default = null
}

locals {
  dotfiles_uri = var.dotfiles_uri != null ? var.dotfiles_uri : data.coder_parameter.dotfiles_uri[0].value
  user         = var.user != null ? var.user : ""
}
```

currently, `var.user`'s default is `null` and I use a ternary operator to send an empty string (`""`) if it's null because `templatefile()` doesn't take null arguments.
I think this is ugly, but while I could set `var.user'`s default to an empty string (`""`), I think it doesn't make sense that the default is not `null` if nothing was passed.

---

`run.sh`
```sh
if [ -n "$${DOTFILES_URI// }" ]; then
  if [ -z "$DOTFILES_USER" ]; then
    DOTFILES_USER="$USER"
  fi
fi
```

In this section, I use `[ -z "$DOTFILES_USER" ]` when I could theoretically use `[ "$${DOTFILES_USER// }" ]` to be more coherent, but I only got it working in my shell and never in the script and ran out of nerve, is it fine if it stays like this ?
Furthemore, I think that the following can look quite confusing :

```sh
if [ -n "$${DOTFILES_URI// }" ]; then
  if [ "$${DOTFILES_USER// }" ]; then
    DOTFILES_USER="$USER"
  fi
fi
```

mafredri:
> Currently, `var.user`'s default is `null` and I use a ternary operator to send an empty string (`""`) to `templatefile()` if it's null because `templatefile()` doesn't take null arguments. I think this is ugly, but while I could set `var.user'`s default to an empty string (`""`), I think it doesn't make sense that the default is not `null` if nothing was passed.

Hmm, defaulting to the empty string should be fine, I don't think null conveys any additional information in this instance. However, perhaps this is a problem for this implementation.

Ultimately, for this to work as expected, we would need the output `dotfiles_uri` from module A to always be a non-empty string. Otherwise module B can't determine whether or not it has been defined.

Maybe we need to scrap the idea? 😔

---

> In this section, I use `[ -z "$DOTFILES_USER" ]` when I could theoretically use `[ "$${DOTFILES_USER// }" ]` to be more coherent, but I only got it working in my shell and never in the script and ran out of nerve, is it fine if it stays like this ? Furthemore, I think that the following can look quite confusing :

My recommendation would be to clean up the input in Terraform and keep the shell script simple(r).

```tf
  script = templatefile("${path.module}/run.sh", {
    DOTFILES_URI : trimspace(local.dotfiles_uri),
    DOTFILES_USER : trimspace(local.user)
  })
```

This way, we don't need `${...// }`.

```shell
if [ -n "$DOTFILES_URI" ]; then
  if [ -z "$DOTFILES_USER" ]; then
    DOTFILES_USER="$USER"
  fi
  # ...
```

matifali (dismissed):
LGTM. Thanks @phorcys420

phorcys420:
> Ultimately, for this to work as expected, we would need the output `dotfiles_uri` from module A to always be a non-empty string. Otherwise module B can't determine whether or not it has been defined.

I don't see the issue, because this scenario where an empty string is returned from module A only happens if the end user didn't provide any dotfiles URI or if we set `dotfiles_uri = ""`, therefore we don't have to apply anything in module B either.
Could you elaborate a bit ? I really don't see how this can cause an issue.

matifali:
> f var.dotfiles_uri == null, then module A's output.dotfiles_uri value will be the coder_parameter's value.

And this can empty as dotfiles are the optional user can leave the coder_aparmeter empty.

phorcys420:
> > f var.dotfiles_uri == null, then module A's output.dotfiles_uri value will be the coder_parameter's value.
> 
> And this can empty as dotfiles are the optional user can leave the coder_aparmeter empty.

yes but then nothing is done with the other module either, it will not error just not do anything.

phorcys420:
hey all, i see this is becoming stale, have you got the time to take a look @mafredri ?

dotfiles/run.sh
```diff
@@ -0,0 +1,21 @@
...
+  echo "✨ Applying dotfiles for user $DOTFILES_USER"
+
+  if [ "$DOTFILES_USER" = "$USER" ]; then
+    coder dotfiles "$DOTFILES_URI" -y 2>&1 | tee -a ~/.dotfiles.log
+  else
+    # The "eval echo ~$USER" part is used to dynamically get the home directory of the user, see https://superuser.com/a/484280
+    # sudo -u coder sh -c 'eval echo ~$USER' -> "/home/coder"                                                                                               ~/projects/coder-modules
+    # sudo sh -c 'eval echo ~$USER'          -> "/root"
+
+    sudo -u "$DOTFILES_USER" sh -c "$(which coder) dotfiles \"$DOTFILES_URI\" -y 2>&1 | tee -a $(eval echo ~$USER)/.dotfiles.log"
```
mafredri, line 19: The escaping here is going to be gnarly, as is it'll return the current users directory, not the target users:

```suggestion
    sudo -u "$DOTFILES_USER" sh -c "'$(which coder)' dotfiles '$DOTFILES_URI' -y 2>&1; echo tee \""'$(eval echo ~$USER)'"\"/.dotfiles.log"
```

Added safety `'`s as well to prevent splitting.

dotfiles/README.md
```diff
@@ -9,12 +9,57 @@ tags: [helper]
...
+```tf
+module "dotfiles" {
+  source   = "registry.coder.com/modules/dotfiles/coder"
+  version  = "1.0.0"
+  agent_id = coder_agent.example.id
+  user     = "root"
+}
+```
+
+### Apply dotfiles as the current user and root (only works if sudo is passwordless)
```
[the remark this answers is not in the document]
phorcys420: hey, I don't think you can do `sudo -u <other user>` when sudo is not passwordless

dotfiles/run.sh
phorcys420: When I wrote the script, I was using `$USER` instead of `$DOTFILES_USER` but quickly realized that was a mistake.
So I only forgot to replace the variable.

This is what I meant to do, do you see any issues with doing it this way ?
```
sudo -u root sh -c "'$(which coder)' dotfiles '$DOTFILES_URI' -y 2>&1 | tee '$(eval echo ~$DOTFILES_USER)/.dotfiles.log'"
```

---

```sh
coder@env> sudo -u root sh -c "'$(which coder)' dotfiles '$DOTFILES_URI' -y 2>&1 | tee '$(eval echo ~$DOTFILES_USER)/.dotfiles.log'"
Checking if dotfiles repository already exists...
Found dotfiles repository at /root/.config/coderv2/dotfiles
warning: redirecting to [...]
Already up to date.
Running setup.sh...
[+] Setting up dotfiles...
[+] Absolute path: /root/.config/coderv2/dotfiles
[+] Changing default shell...
[+] Changing PAM policy for chsh
[+] Setting default shell
[+] Reverting PAM policy for chsh
[+] Could not change default shell (99)
Dotfiles installation complete.
coder@env> sudo -u root stat /root/.dotfiles.log                                                                                 ~
  File: /root/.dotfiles.log
  Size: 488             Blocks: 8          IO Block: 4096   regular file
Device: 10008dh/1048717d        Inode: 266456      Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2024-03-07 17:29:59.822682570 +0000
Modify: 2024-03-07 17:29:48.174934810 +0000
Change: 2024-03-07 17:29:48.174934810 +0000
 Birth: 2024-03-07 17:29:47.694945204 +0000
```

dotfiles/main.tf
```diff
@@ -14,7 +14,23 @@ variable "agent_id" {
   description = "The ID of a Coder agent."
 }
 
+variable "dotfiles_uri" {
+  type        = string
+  description = "The URL to a dotfiles repository. (optional)"
```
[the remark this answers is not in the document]
phorcys420: @mafredri should we mention it in the description or in the README, or in both ?

dotfiles/run.sh
mafredri: That looks fine, you could do `~"$DOTFILES_USER"` (vs `~$DOTFILES_USER`) just to be on the safe side.

dotfiles/main.tf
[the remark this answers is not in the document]
mafredri: I think it's enough to mention it in the description here. Up to you though.

dotfiles/README.md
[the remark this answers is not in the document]
mafredri: With this chance I'm referring to two entities, the "user" dotfiles, and the "root" dotfiles. The parens was meant to suggest that "root" dotfiles only work if sudo is passwordless.

In my previous suggestion we already mention that sudo always has to be passwordless when applying for "another user".

How about this rewrite instead?

```suggestion
### Apply the same dotfiles as the current user and root (the root dotfiles can only be applied if sudo is passwordless)
```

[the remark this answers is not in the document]
phorcys420: oh i see what you mean now, yes i do like this other rewrite better

dotfiles/main.tf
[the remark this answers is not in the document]
phorcys420: a2bdfef

dotfiles/run.sh
phorcys420: Since we have to escape anyways, I opted for single instead of double quotes just to be on the safer side.
[63c0712](https://github.com/phorcys420/coder-modules/commit/63c0712cd4a67a260b9e9aee07e83004a6839dee)

dotfiles/run.sh
```diff
@@ -0,0 +1,21 @@
...
+  fi
+
+  echo "✨ Applying dotfiles for user $DOTFILES_USER"
+
+  if [ "$DOTFILES_USER" = "$USER" ]; then
+    coder dotfiles "$DOTFILES_URI" -y 2>&1 | tee ~/.dotfiles.log
+  else
+    # The "eval echo ~$DOTFILES_USER" part is used to dynamically get the home directory of the user, see https://superuser.com/a/484280
+    # sudo -u coder sh -c 'eval echo ~$DOTFILES_USER' -> "/home/coder"
+    # sudo sh -c 'eval echo ~$DOTFILES_USER'          -> "/root"
```
mafredri, line 17: Why do these look different? Both can just run as the normal user, no sudo involved. The variable name being the same is confusing.

```suggestion
    # eval echo ~$USER'                   -> "/home/coder"
    # eval echo ~${DOTFILES_USER:-root}'  -> "/root"
```

dotfiles/run.sh
```diff
@@ -0,0 +1,21 @@
...
+
+    sudo -u "$DOTFILES_USER" sh -c "$(which coder) dotfiles \"$DOTFILES_URI\" -y 2>&1 | tee -a $(eval echo ~\'$DOTFILES_USER\')/.dotfiles.log"
```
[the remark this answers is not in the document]
phorcys420: I did intend to run the commands as the specified use at first, that's why past commits used `$USER` instead of `$DOTFILES_USER`.

But it is indeed confusing now that it's not required.

dotfiles/run.sh
phorcys420: I changed the variable the other day because I got mixed up but the examples were actually meant to use `$USER` and since we use single quotes that command would've been run as the current user within the sudo command, so `$USER` was correct in the previous commits.

Now the way we resolve the home changed though so we can indeed change it.

phorcys420: 6ffc79f

dotfiles/run.sh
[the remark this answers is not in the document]
phorcys420: 6ffc79f

phorcys420:
hey!
sorry for letting this PR become stale, I went ahead and applied the changes you suggested.

dotfiles/run.sh
phorcys420: Actually, this was causing linting issues so I changed it to the following, which is clearer anyways in my opinion.

```bash
  # The `eval echo ~"$DOTFILES_USER"` part is used to dynamically get the home directory of the user, see https://superuser.com/a/484280
  # eval echo ~coder -> "/home/coder"
  # eval echo ~root  -> "/root"
``

0b8fe33

**Comment by matifali** [comment]
@phorcys420 can you rebase it?

**Review by matifali — APPROVED** [review]
@mafredi I am approving it but wait for you to merge in case if you want to have a final review.

**Comment by phorcys420** [comment]
hey @matifali, just resolved the merge conflict, there was an error in the docs where the variable name was `default_dotfiles_repo` instead of `default_dotfiles_uri` so I corrected that.

Also, does it make sense that there is a `dotfiles_default_uri` output? I don't see how this could be useful.
```tf
output "dotfiles_default_uri" {
  description = "Dotfiles Default URI"
  value       = var.default_dotfiles_uri
}
```

phorcys420:
Another thing, could you review the README.md to see if it makes sense to you?
I think the section about default dotfiles URI might be better somewhere else, but I don't really know.

matifali:
@phorcys420 having an output doesn't do any harm but could be helpful for some users, so I did not remove it. I agree that to doesn't add much value as its not anything computed but a direct output from an input. 

Let's remove that.

phorcys420:
@matifali do you want me to remove it in this PR?

dotfiles/main.tf
```diff
@@ -30,25 +46,28 @@ data "coder_parameter" "dotfiles_uri" {
...
-  value       = data.coder_parameter.dotfiles_uri.value
+  value       = local.dotfiles_uri
 }
 
 output "dotfiles_default_uri" {
   description = "Dotfiles Default URI"
   value       = var.default_dotfiles_uri
-}
\ No newline at end of file
+}
```
matifali, line 73: Let us remove it.

dotfiles/README.md
```diff
@@ -19,15 +23,55 @@ module "dotfiles" {
...
-  source                = "registry.coder.com/modules/dotfiles/coder"
-  version               = "1.0.12"
-  agent_id              = coder_agent.example.id
-  default_dotfiles_repo = "https://github.com/coder/dotfiles"
+  source               = "registry.coder.com/modules/dotfiles/coder"
+  version              = "1.0.12"
+  agent_id             = coder_agent.example.id
+  default_dotfiles_uri = "https://github.com/coder/dotfiles"
 }
 ```
```
matifali, line 77: I think this is ok.

matifali (commented):
done

matifali:
@phorcys420 We can do it in a separate PR and keep the changes related in this PR.

The new PR can fix the typo and remove output.

phorcys420:
@matifali so should I apply the change you suggested?
since you said you want to do it in another PR?

matifali:
#225 does that but we need to wait for someone to approve that :D

phorcys420:
oh okay, but I already fixed the typo in this branch @matifali, do you want me to unfix it to keep the changes related to the PR? 😄

matifali:
@phorcys420 Yes, its better to do ina separate PR.

phorcys420:
I have tested the module again after all the changes and I can confirm that it still works as expected.

matifali:
@phorcys420 Could you rebase it again. :D

phorcys420:
> @phorcys420 Could you rebase it again. :D

will do!

dotfiles/main.tf
```diff
@@ -16,17 +16,32 @@ variable "agent_id" {
...
+  description = "The default dotfiles URI if the workspace user does not provide one"
   default     = ""
 }
 
+variable "dotfiles_uri" {
+  type        = string
+  description = "The URL to a dotfiles repository. (optional, when set, the user isn't prompted for their dotfiles)"
+
+  default = null
+}
```
matifali, line 28: I think we should only have one of these

PR #134, opened 2024-01-27
chore: add prettier terraform formatting in markdown files

mafredri:
This PR enables Terraform formatting in markdown files (via `bun fmt`). Since the `prettier-plugin-terraform-formatter` plugin doesn't register `hcl` we also rename all code-blocks to `tf`, atlernatively we could patch the plugin but I think this is fine.

- chore: add prettier terraform formatting in markdown files
- rename `hcl` codeblocks to `tf` to enable formatter
- bun fmt
- fix invalid tf
- bun fmt

### DIFF
--- a/.sample/README.md
+++ b/.sample/README.md
@@ -11,7 +11,7 @@ tags: [helper]
 
 <!-- Describes what this module does -->
 
-```hcl
+```tf
 module "MODULE_NAME" {
   source = "registry.coder.com/modules/MODULE_NAME/coder"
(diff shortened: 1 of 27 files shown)


matifali (approved):
Awesome 🎉

matifali:
Should we add a lint check to make sure there is no `hcl` codeblocks?

mafredri:
> Should we add a lint check to make sure there is no `hcl` codeblocks?

That's a good idea 👍🏻

PR #135, opened 2024-01-27
chore: lint for tf/hcl blocks

mafredri:
This PR lints that there is at least one `tf` block and no `hcl` blocks in the READMEs.

Ref: https://github.com/coder/modules/pull/134#issuecomment-1913148604

### DIFF
--- a/lint.ts
+++ b/lint.ts
@@ -13,9 +13,39 @@ let badExit = false;
 // error reports an error to the console and sets badExit to true
 // so that the process will exit with a non-zero exit code.
 const error = (...data: any[]) => {
-    console.error(...data);
-    badExit = true;
-}
+  console.error(...data);
+  badExit = true;
+};
+
+const verifyCodeBlocks = (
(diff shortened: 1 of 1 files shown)


lint.ts
```diff
@@ -13,9 +13,39 @@ let badExit = false;
...
+const verifyCodeBlocks = (
+  tokens: marked.Token[],
+  res = {
+    codeIsTF: false,
+    codeIsHCL: false,
+  }
+) => {
+  for (const token of tokens) {
+    // Check in-depth.
+    if (token.type === "list") {
```
mafredri, line 29: Review: There may be other in-depth cases that are unhandled, but this was present in `slackme` README, at least.

matifali:
I also added this in #128 
your choice :D

mafredri:
Feel free to import these changes into your PR if you want, there's a difference in that this PR lints all code blocks, even after h1 and within lists.

matifali (approved):

PR #140, opened 2024-01-31
Update Hashicorp vault modules

matifali:
This pull request adds a new module for authenticating with Hashicorp Vault using a token. The module allows users to configure the Vault address, token, and version of the Vault CLI to install.


vault-token/run.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Fetch the latest version of Vault if INSTALL_VERSION is 'latest'
+if [ "$${INSTALL_VERSION}" = "latest" ]; then
+  LATEST_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -oP 'vault/\K[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n 1)
+  printf "Latest version of Vault is %s.\n\n" "$${LATEST_VERSION}"
+  if [ -z "$${LATEST_VERSION}" ]; then
+    printf "Failed to determine the latest Vault version.\n"
+    exit 1
+  fi
+  VERSION=$${LATEST_VERSION}
```
mafredri, line 40:
```suggestion
  INSTALL_VERSION=$${LATEST_VERSION}
```

vault-token/run.sh
```diff
@@ -0,0 +1,82 @@
...
+    installation_needed=0
+  fi
+fi
+
+if [ $${installation_needed} -eq 1 ]; then
+  # Download and install Vault
+  if [ -z "$${CURRENT_VERSION}" ]; then
+    printf "Installing Vault CLI ...\n\n"
+  else
+    printf "Upgrading Vault CLI from version %s to %s ...\n\n" "$${CURRENT_VERSION}" "$${VERSION}"
```
mafredri, line 58:
```suggestion
    printf "Upgrading Vault CLI from version %s to %s ...\n\n" "$${CURRENT_VERSION}" "$${INSTALL_VERSION}"
```

vault-token/run.sh
```diff
@@ -0,0 +1,82 @@
...
+  fi
+  fetch vault.zip "https://releases.hashicorp.com/vault/$${VERSION}/vault_$${VERSION}_linux_amd64.zip"
```
mafredri, line 60:
```suggestion
  fetch vault.zip "https://releases.hashicorp.com/vault/$${INSTALL_VERSION}/vault_$${INSTALL_VERSION}_linux_amd64.zip"
```

vault-token/run.sh
```diff
@@ -0,0 +1,82 @@
...
+installation_needed=1
+if command -v vault > /dev/null 2>&1; then
+  CURRENT_VERSION=$(vault version | grep -oE '[0-9]+\.[0-9]+\.[0-9]+')
+  if [ "$${CURRENT_VERSION}" = "$${INSTALL_VERSION}" ]; then
+    printf "Vault version %s is already installed and up-to-date.\n\n" "$${CURRENT_VERSION}"
+    installation_needed=0
+  fi
+fi
+
+if [ $${installation_needed} -eq 1 ]; then
```
mafredri, line 53: It might be a good idea to ensure we're in a temporary directory when doing this.

```shell
install() {
  # move the code inside func ... no exit 1, just return 1
}

TMP=$(mktemp -d)
pushd "$${TMP}"
if ! install; then
  printf failed
fi
popd "$${TMP}"
rm -rf "$${TMP}"
```

(By moving the logic we can also ensure that the cleanup always runs.)

vault-token/run.sh
```diff
@@ -0,0 +1,82 @@
...
+  else
+    printf "Upgrading Vault CLI from version %s to %s ...\n\n" "$${CURRENT_VERSION}" "$${VERSION}"
+  fi
+  fetch vault.zip "https://releases.hashicorp.com/vault/$${VERSION}/vault_$${VERSION}_linux_amd64.zip"
+  if [ $? -ne 0 ]; then
+    printf "Failed to download Vault.\n"
+    exit 1
+  fi
+  unzip vault.zip
+  if [ $? -ne 0 ]; then
```
mafredri, line 66: These can be simplified (see other spots too):

```suggestion
  if ! unzip vault.zip; then
```

vault-token/run.sh
```diff
@@ -0,0 +1,82 @@
...
+    printf "Failed to unzip Vault.\n"
+    exit 1
+  fi
+  rm vault.zip
+  if sudo mv vault /usr/local/bin/vault 2> /dev/null; then
+    printf "Vault installed successfully!\n\n"
+  else
+    mkdir -p ~/.local/bin
+    mv vault ~/.local/bin/vault
+    if [ ! -f ~/.local/bin/vault ]; then
```
mafredri, line 76:
```suggestion
    if ! mv vault ~/.local/bin/vault; then
```

vault-token/run.sh
matifali: I am also considering reusing this vault CLI installation code across vault modules. Nested modules?

vault-github/run.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
mafredri, line 1: This is probably my bad for suggesting the use of `pushd`/`popd` previously, forgot it's a Bashism. I think it's better to keep all modules limited to `sh` functionality, that way we have the best compatibility, sometimes Bash may not be present.

The `pushd`/`popd` commands can be replaced by `cd "..."`/`cd -`, or a subshell form like this:

```sh
if ! (
	cd /tmp/...
	install
); then
	echo install fail
fi
# no need to cd here because cd was contained in subshell ( ... ).
```

vault-github/run.sh
```diff
@@ -31,57 +31,65 @@ unzip() {
...
+    else
+      mkdir -p ~/.local/bin
+      if ! mv vault ~/.local/bin/vault; then
+        printf "Failed to move Vault to local bin.\n"
+        return 1
+      fi
+      printf "Please add ~/.local/bin to your PATH to use vault CLI.\n"
     fi
-    printf "Please add ~/.local/bin to your PATH to use vault CLI.\n"
   fi
```
mafredri, line 83: Suggestion: With these types of functions, it's IMO a good practice to do an explicit return 0 (OK) at the end because is makes the intent clear, i.e. if the code reaches here we're all good. Otherwise we are at the mercy of whatever command or expression was evaluated last.

mafredri (approved):
Looks good! Just a couple of minor suggestions.

vault-github/run.sh
matifali: Sure. I will return 0.

vault-github/run.sh
```diff
@@ -31,57 +31,68 @@ unzip() {
...
+  return 0
+}
+
+TMP=$(mktemp -d)
+if ! (
+  cd $TMP
```
mafredri, line 89: It may be rare, but tmpdir could contain spaces
```suggestion
  cd "$TMP"
```

vault-github/run.sh
```diff
@@ -31,57 +31,68 @@ unzip() {
...
+  install
+); then
+  echo "Failed to install Vault CLI."
+  exit 1
 fi
+rm -rf $TMP
```
mafredri, line 95:
```suggestion
rm -rf "$TMP"
```

vault-token/run.sh
```diff
@@ -0,0 +1,93 @@
...
+      fi
+      printf "Please add ~/.local/bin to your PATH to use vault CLI.\n"
+    fi
+  fi
+  return 0
+}
+
+TMP=$(mktemp -d)
+if ! (
+  cd $TMP
```
mafredri, line 87:
```suggestion
  cd "$TMP"
```

vault-token/run.sh
```diff
@@ -0,0 +1,93 @@
...
+  install
+); then
+  echo "Failed to install Vault CLI."
+  exit 1
+fi
+rm -rf $TMP
```
mafredri, line 93:
```suggestion
rm -rf "$TMP"
```

mafredri (approved):

PR #141, opened 2024-02-02
feat!(git-config): use full name for git configuration

### LINKED ISSUE #126: add Full username to git-config module
After https://github.com/coder/terraform-provider-coder/pull/181, a Full Name field is available under `data.coder_workspace.me.owner_name`. We can use this to set `git config --user.name ${data.coder_workspace.me.owner_name}` in [git-config](https://registry.coder.com/modules/git-config) module.

This needs a minimum version of Coder [`v2.7.0`](https://github.com/coder/coder/releases/tag/v2.7.0) and `coder-terraform-provider` [`v0.13.0`](https://github.com/coder/terraform-provider-coder/releases/tag/v0.13.0). We need to check if the Coder is on the correct version.

matifali:
closes #126


git-config/main.tf
```diff
@@ -44,18 +44,31 @@ data "coder_parameter" "username" {
   name         = "username"
   type         = "string"
   default      = ""
-  description  = "Git user.name to be used for commits. Leave empty to default to Coder username."
+  description  = "Git user.name to be used for commits. Leave empty to default to Coder user's Full Name."
   display_name = "Git config user.name"
```
stirby, line 48: Do you think this display name is most intuitive?

Or should it be replaced with something like "Full name for Git config"?

git-config/main.tf
```diff
@@ -44,18 +44,31 @@ data "coder_parameter" "username" {
...
+
+resource "coder_env" "git_commmiter_name" {
+  agent_id = var.agent_id
+  name     = "GIT_COMMITTER_NAME"
+  value    = coalesce(try(data.coder_parameter.username[0].value, ""), data.coder_workspace.me.owner_name, data.coder_workspace.me.owner)
+}
+
+resource "coder_env" "git_author_email" {
+  agent_id = var.agent_id
+  name     = "GIT_AUTHOR_EMAIL"
```
stirby, line 66: Smart to set these in the env rather than through git CLI. 👍

stirby (approved):
nice

git-config/main.tf
matifali: Hmm yeah. We can replace it.

mtojek (commented):
Looks like the CI complains about syntax?

matifali:
@mtojek yes. Our tests run `terraform apply` and for some reason `data.coder_workspace.me.owner_email` is being seen as an empty string by `coalesce` function. 😕

matifali:
As this is not a critical module. I am removing the tests. cc: @stirby

PR #144, opened 2024-02-06
Add Hashicorp Vault Secrets Integration module

matifali:


hcp-vault-secrets/README.md
```diff
@@ -0,0 +1,68 @@
...
+  source   = "registry.coder.com/modules/hcp-vault-secrets/coder"
+  version  = "1.0.3"
+  agent_id = coder_agent.example.id
+  app_name = "demo-app"
+}
+```
+
+## Fetch Selective Secrets
+
+To fetch selective secrets from the HCP Vault Secrets app, set the `secrets` input.
```
mafredri, line 43: I was going to suggest "selective" -> "select", but Open AI says it's fine and maybe slightly more clear with "selective" 😄

hcp-vault-secrets/main.tf
```diff
@@ -0,0 +1,67 @@
...
+data "hcp_vault_secrets_app" "secrets" {
+  app_name = var.app_name
+}
+
+resource "coder_env" "hvs_secrets" {
+  # https://support.hashicorp.com/hc/en-us/articles/4538432032787-Variable-has-a-sensitive-value-and-cannot-be-used-as-for-each-arguments
+  for_each = var.secrets != null ? toset(var.secrets) : nonsensitive(toset(keys(data.hcp_vault_secrets_app.secrets.secrets)))
+  agent_id = var.agent_id
+  name     = each.key
+  value    = data.hcp_vault_secrets_app.secrets.secrets[each.key]
```
mafredri, line 66: Can't you use `each.value` here instead?

mafredri (approved):

hcp-vault-secrets/README.md
matifali: So Open AI or Mathias?

hcp-vault-secrets/main.tf
matifali: I am looping over the keys only so they do not have a value. If we loop over the full map `data.hcp_vault_secrets_app.secrets.secrets` then we can use `each.value` but that would require converting secret values to be nonsensitive too.

hcp-vault-secrets/README.md
```diff
@@ -0,0 +1,68 @@
...
+}
+```
+
+## Configuration
+
+To configure the HCP Vault Secrets module, you must create an HCP Service Principal from the HCP Vault Secrets app in the HCP console. This will give you the `HCP_CLIENT_ID` and `HCP_CLIENT_SECRET` that you need to authenticate with HCP Vault Secrets. See the [HCP Vault Secrets documentation](https://developer.hashicorp.com/hcp/docs/vault-secrets) for more information.
+
+## Fetch All Secrets
+
+To fetch all secrets from the HCP Vault Secrets app, skip the `secrets` input.
```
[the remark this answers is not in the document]
matifali: Hmm. It's subjective. As per my experience, HVS is designed to create one app per project/repo, and injecting all should be the default. if someone is using a single app for all their secrets of multiple repos/projects, then it's not a good default.

hcp-vault-secrets/main.tf
mafredri: Ah yeah, that makes sense 👍🏻

hcp-vault-secrets/README.md
mafredri: I don't trust myself, nor do I trust AI. It's a tough problem for sure 😂

hcp-vault-secrets/README.md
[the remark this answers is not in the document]
matifali: I can add an additional input `fetch_all` or `all` in this case which will ignore `secrets` and fetch everything with default set to `false`.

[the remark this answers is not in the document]
mafredri: I think it's fine to leave like this if the intent is to create one app per project 👍🏻

matifali:
I have not added tests. As our tests run `terraform apply` which require real credentials for the provider we use in the module.

Any thoughts? @mafredri

@kylecarbs helped in writing a [fake API server](https://github.com/coder/modules/blob/acab6437bca3201d33604ddd1add550ca6d8c5b4/jfrog-token/main.test.ts#L14) for the `jfrog-token` module.

PR #150, opened 2024-02-08
Add build numbers and versions to jetbrains gateway module

### LINKED ISSUE #145: Hardcoded IDE versions in locals should be refactored into variables
Eg, in https://github.com/coder/modules/blob/acab6437bca3201d33604ddd1add550ca6d8c5b4/jetbrains-gateway/main.tf#L62 the cal-versioned dates should be in a a variable and then included in the local diffs so that downstream template operators can control which versions appear as linked the in UI.

asvoboda:
Resolves #145

### DIFF
--- a/jetbrains-gateway/main.tf
+++ b/jetbrains-gateway/main.tf
@@ -30,17 +30,68 @@ variable "default" {
   description = "Default IDE"
 }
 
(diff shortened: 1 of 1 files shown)


code-asher (approved):
Looks good to me!

PR #152, opened 2024-02-09
fix(git-commit-signing): fix SSH key permissions

### LINKED ISSUE #149: Git commit signing Permission denied
Hi there,

I use the [Git commit signing](https://registry.coder.com/modules/git-commit-signing?tab=variables) module and every time my workspace starts, I get this error:
```sh
sh: 26: cannot create /home/User/.ssh/git-commit-signing/coder.pub: Permission denied
sh: 30: cannot create /home/User/.ssh/git-commit-signing/coder: Permission denied
```

It seems the keys have the following permission:
```
-r-------- 1 User User 388 Feb  7 23:33 coder
-r-------- 1 User User 82 Feb  7 23:33 coder.pub
```

Any idea how to fix this? Thanks!

phorcys420:
Closes #149.
in Linux it's fine to modify a file's permissions that only has read permissions if you're the owner, but you still can't write to them, so if the home folder of the user is persisted, then it'll cause issues on every startup, this PR makes the files writable by the owner.

`~/.ssh/git-commit-signing/coder` goes from `400` to `600`
`~/.ssh/git-commit-signing/coder.pub` goes from `400` to `644`

these are the default values when using `ssh-keygen`, I don't know why I didn't use those permissions when writing the script.

### DIFF
--- a/git-commit-signing/run.sh
+++ b/git-commit-signing/run.sh
@@ -31,8 +31,8 @@ jq --raw-output ".private_key" > ~/.ssh/git-commit-signing/coder << EOF
 $ssh_key
 EOF
 
-chmod -R 400 ~/.ssh/git-commit-signing/coder
-chmod -R 400 ~/.ssh/git-commit-signing/coder.pub
+chmod -R 600 ~/.ssh/git-commit-signing/coder
+chmod -R 644 ~/.ssh/git-commit-signing/coder.pub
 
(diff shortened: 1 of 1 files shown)


phorcys420:
cc: @matifali

matifali (approved):

matifali:
Closes #149

PR #154, opened 2024-02-13
Add extensions support for vscode-web

matifali:
This pull request adds support for installing extensions in vscode-web.

### DIFF
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -53,19 +53,41 @@ variable "log_path" {
   default     = "/tmp/vscode-web.log"
 }
 
-variable "install_dir" {
+variable "install_prefix" {
   type        = string
-  description = "The directory to install VS Code CLI"
-  default     = "/tmp/vscode-cli"
+  description = "The prefix to install vscode-web to."
+  default     = "/tmp/vscode-server"
+}
+
+variable "extensions" {
+  type        = list(string)
+  description = "A list of extensions to install."
+  default     = []
 }
 
 variable "accept_license" {
   type        = bool
-  description = "Accept the VS Code license. https://code.visualstudio.com/license"
+  description = "Accept the VS Code Server license. https://code.visualstudio.com/license/server"
   default     = false
   validation {
     condition     = var.accept_license == true
     error_message = "You must accept the VS Code license agreement by setting accept_license=true."
+  }
+}
+
+variable "extensions" {
+  type        = list(string)
+  description = "A list of extensions to install."
+  default     = []
+}
+
+variable "telemetry_level" {
+  type        = string
+  description = "Set the telemetry level for VS Code Web."
+  default     = "error"
+  validation {
+    condition     = var.telemetry_level == "off" || var.telemetry_level == "crash" || var.telemetry_level == "error" || var.telemetry_level == "all"
+    error_message = "Incorrect value. Please set either 'off', 'crash', 'error', or 'all'."
   }
 }
 
@@ -76,7 +98,9 @@ resource "coder_script" "vscode-web" {
   script = templatefile("${path.module}/run.sh", {
     PORT : var.port,
     LOG_PATH : var.log_path,
-    INSTALL_DIR : var.install_dir,
+    INSTALL_PREFIX : var.install_prefix,
+    EXTENSIONS : join(",", var.extensions),
+    TELEMETRY_LEVEL : var.telemetry_level,
   })
   run_on_start = true
 }
--- a/vscode-web/run.sh
+++ b/vscode-web/run.sh
@@ -1,21 +1,52 @@
 #!/usr/bin/env sh
 
 BOLD='\033[0;1m'
+EXTENSIONS=("${EXTENSIONS}")
 
-# Create install directory if it doesn't exist
-mkdir -p ${INSTALL_DIR}
+# Create install prefix
+mkdir -p ${INSTALL_PREFIX}
 
-printf "$${BOLD}Installing vscode-cli!\n"
+printf "$${BOLD}Installing Microsoft Visual Studio Code Server!\n"
 
-# Download and extract code-cli tarball
-output=$(curl -Lk 'https://code.visualstudio.com/sha/download?build=stable&os=cli-alpine-x64' --output vscode_cli.tar.gz && tar -xf vscode_cli.tar.gz -C ${INSTALL_DIR} && rm vscode_cli.tar.gz)
+# Download and extract vscode-server
+ARCH=$(uname -m)
+case "$ARCH" in
+  x86_64) ARCH="x64" ;;
+  aarch64) ARCH="arm64" ;;
+  *)
+    echo "Unsupported architecture"
+    exit 1
+    ;;
+esac
+
+HASH=$(curl https://update.code.visualstudio.com/api/commits/stable/server-linux-$ARCH-web | cut -d '"' -f 2)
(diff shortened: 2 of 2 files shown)


vscode-web/main.tf
```diff
@@ -53,30 +53,48 @@ variable "log_path" {
   default     = "/tmp/vscode-web.log"
 }
 
-variable "install_dir" {
+variable "install_prefix" {
   type        = string
-  description = "The directory to install VS Code CLI"
-  default     = "/tmp/vscode-cli"
+  description = "The prefix to install vscode-web to."
+  default     = "/tmp/vscode-server"
```
mafredri, line 59:
```suggestion
  default     = "/tmp/vscode-web"
```

We should probably attempt naming consistency between our examples and default values.

vscode-web/run.sh
```diff
@@ -1,21 +1,49 @@
...
-printf "$${BOLD}Installing vscode-cli!\n"
+printf "$${BOLD}Installing Microsoft Visual Studio Code Server!\n"
 
-# Download and extract code-cli tarball
-output=$(curl -Lk 'https://code.visualstudio.com/sha/download?build=stable&os=cli-alpine-x64' --output vscode_cli.tar.gz && tar -xf vscode_cli.tar.gz -C ${INSTALL_DIR} && rm vscode_cli.tar.gz)
+# Download and extract vscode-server
+ARCH=$(uname -m)
+case "$ARCH" in
+  x86_64) ARCH="x64" ;;
+  aarch64) ARCH="arm64" ;;
```
mafredri, line 15: We could also support 32bit ARM (`ARCH=armhf`), not sure what `uname -m` would print in this case, though.

vscode-web/run.sh
```diff
@@ -1,21 +1,49 @@
...
+  *)
+    echo "Unsupported architecture"
+    exit 1
+    ;;
+esac
+
+HASH=$(curl https://update.code.visualstudio.com/api/commits/stable/server-linux-$ARCH-web | cut -d '"' -f 2)
+outut=$(curl -sL https://vscode.download.prss.microsoft.com/dbazure/download/stable/$HASH/vscode-server-linux-$ARCH-web.tar.gz | tar -xz -C ${INSTALL_PREFIX} --strip-components 1 && rm -rf vscode-server-linux-$ARCH-web.tar.gz)
```
mafredri, line 23: Typo:

```suggestion
output=$(curl -sL https://vscode.download.prss.microsoft.com/dbazure/download/stable/$HASH/vscode-server-linux-$ARCH-web.tar.gz | tar -xz -C ${INSTALL_PREFIX} --strip-components 1)
```

Also, `rm` seems a remnant of a previous iteration(?), it's currently not saved to disk AFAICT.

vscode-web/run.sh
```diff
@@ -1,21 +1,49 @@
...
 
 if [ $? -ne 0 ]; then
-  echo "Failed to install vscode-cli: $output"
+  echo "Failed to install vscode-server: $output"
   exit 1
 fi
-printf "🥳 vscode-cli has been installed.\n\n"
+echo "🥳 vscode-server has been installed in ${INSTALL_PREFIX}\n"
```
mafredri: Nit: Inconsistent terminology, to says `Installing Microsoft Visual Studio Code Server`, but later we call it just `vscode-server`.

vscode-web/run.sh
```diff
@@ -1,21 +1,49 @@
...
+for extension in "$${EXTENSIONLIST[@]}"; do
+  if [ -z "$extension" ]; then
+    continue
+  fi
+  printf "🧩 Installing extension $${CODE}$extension$${RESET}...\n"
+  output=$($VSCODE_SERVER --install-extension "$extension" --force)
+  if [ $? -ne 0 ]; then
+    echo "Failed to install extension: $extension: $output"
+    exit 1
+  fi
```
mafredri, line 44: Given:

```sh
EXTENSIONS="foo,bar,baz"
```

I would suggest the following posix way, so we don't need to rely on bash/arrays.

```suggestion
extension=
while [ "$${extension}" != "$${EXTENSIONS}" ]; do
  extension=$${EXTENSIONS%%,*}
  EXTENSIONS=$${EXTENSIONS#*,}
  if [ -z "$${extension}" ]; then
    continue
  fi
  printf "🧩 Installing extension $${CODE}$extension$${RESET}...\n"
  output=$("$${VSCODE_SERVER}" --install-extension "$${extension}" --force)
  if [ $? -ne 0 ]; then
    echo "Failed to install extension: $${extension}: $${output}"
    exit 1
  fi
```

vscode-web/run.sh
```diff
@@ -1,21 +1,49 @@
...
+done
 
-echo "👷 Running ${INSTALL_DIR}/bin/code serve-web --port ${PORT} --without-connection-token --accept-server-license-terms in the background..."
+echo "👷 Running ${INSTALL_PREFIX}/bin/code-server"
 echo "Check logs at ${LOG_PATH}!"
-${INSTALL_DIR}/code serve-web --port ${PORT} --without-connection-token --accept-server-license-terms > ${LOG_PATH} 2>&1 &
+${INSTALL_PREFIX}/bin/code-server serve-local --port ${PORT} --accept-server-license-terms serve-local --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} > ${LOG_PATH} 2>&1 &
```
mafredri, line 49: Nit: It might be a good idea to use intermediary shell variables, but in either case they should always be quoted for commands to avoid splitting.

```suggestion
"${INSTALL_PREFIX}/bin/code-server" serve-local --port "${PORT}" --accept-server-license-terms serve-local --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
```

vscode-web/run.sh
mafredri: PS. the `-S` option for `curl`, (`curl -sSL`) would allow errors to be shown, I believe `-s` silences them currently.

matifali: yup I copied this over from my other template. thanks

vscode-web/run.sh
matifali: We can use the `Microsoft Visual Studio Code Server` consistently.

vscode-web/run.sh
matifali: This section is borrowed from [`code-server`](https://github.com/coder/modules/blob/b299f98161a02fe335c7e06ff4c93ce8fc24137a/code-server/run.sh#L28), and I did not try to improve it as code-server is our top downloaded module and well tested. But I agree and we can update.

vscode-web/run.sh
matifali: we can do it if we get any feature requests. currently, I can not access it.

vscode-web/run.sh
matifali: @mafredri, any second thoughts? Otherwise I should also update code-server module.

mafredri (approved):

test.ts
```diff
@@ -42,7 +43,9 @@ export const executeScriptInContainer = async (
 }> => {
   const instance = findResourceInstance(state, "coder_script");
   const id = await runContainer(image);
-  const resp = await execContainer(id, [shell, "-c", instance.script]);
+  // const resp = await execContainer(id, [shell, "-c", instance.script]);
+  command = command + " " + instance.script;
```
mafredri, line 47:
```suggestion
  command = command + "; " + instance.script;
```

matifali:
@mafredri I removed the tests as our modification to run commands was causing many other tests to fail. 
We can add tests again in a separate PR.

PR #156, opened 2024-02-14
fix: fix fetching rc versions of vault cli

matifali:
Fixes where the install script was fetching the RC version

### DIFF
--- a/vault-token/run.sh
+++ b/vault-token/run.sh
@@ -32,7 +32,7 @@ unzip_safe() {
 install() {
   # Fetch the latest version of Vault if INSTALL_VERSION is 'latest'
   if [ "$${INSTALL_VERSION}" = "latest" ]; then
-    LATEST_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -v 'rc' | grep -oP 'vault/\K[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n 1)
+    LATEST_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -v '-rc' | grep -oP 'vault/\K[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n 1)
     printf "Latest version of Vault is %s.\n\n" "$${LATEST_VERSION}"
     if [ -z "$${LATEST_VERSION}" ]; then
       printf "Failed to determine the latest Vault version.\n"
--- a/vault-token/run.sh
+++ b/vault-token/run.sh
@@ -32,7 +32,7 @@ unzip_safe() {
 install() {
   # Fetch the latest version of Vault if INSTALL_VERSION is 'latest'
   if [ "$${INSTALL_VERSION}" = "latest" ]; then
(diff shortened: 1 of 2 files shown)


mtojek (commented):
Don't you want to use [index.json](https://releases.hashicorp.com/vault/index.json) endpoint?

mtojek:
Interesting. Can you shed more light on why? I thought that `jq` format was more human-friendly.

matifali:
This is a module, and if a workspace does not have `jq`, then it will fail. Ideally, we want 0 dependencies if possible, and the module should be useable in any workspace.

@mafredri once suggested having `coder util jq|curl|wget|zip|tar` type of commands so that we do not depend on anything.

mafredri:
> This is a module, and if a workspace does not have `jq`, then it will fail. Ideally, we want 0 dependencies if possible, and the module should be useable in any workspace.
> 
> @mafredri once suggested having `coder util jq|curl|wget|zip|tar` type of commands so that we do not depend on anything.

Yup, this is the way, let's reduce dependencies wherever we can 💪🏻. The `coder util` is a beautiful idea but we also want to make sure we don't bloat the agent binary. Most of this can be done in a very basic way using the stdlib though, so shouldn't add much weight.

Then again, maybe we don't need to bundle it in `coder util`, perhaps we can have module dependencies and we have one that's called `coder/modules/util` that contains a few basic commands, or the ability to download them for the relevant platform.

vault-token/run.sh
```diff
@@ -32,7 +32,7 @@ unzip_safe() {
 install() {
   # Fetch the latest version of Vault if INSTALL_VERSION is 'latest'
   if [ "$${INSTALL_VERSION}" = "latest" ]; then
-    LATEST_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -oP 'vault/\K[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n 1)
+    LATEST_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -v 'rc' | grep -oP 'vault/\K[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n 1)
```
mafredri, line 35:
```suggestion
    LATEST_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -v '-rc' | grep -oP 'vault/\K[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n 1)
```

mafredri (approved):

matifali:
@mafredri @mtojek another option is to use
`curl -s https://releases.hashicorp.com/vault/index.json | grep -o '"[0-9]\+\.[0-9]\+\.[0-9]\+"' | tr -d '"' | grep -v 'rc' | sort -V | tail -n 1`
without using `jq` it also uses the index.json endpoint

mtojek:
> another option is to use

I think I like the first option, curl + regular endpoint + grep. As `index.json` is one-line response, I would parse it without `jq`.

PR #158, opened 2024-02-14
feat(jfrog): add option to customize server id for JFrog CLI

dimberr:


matifali:
Hi @dimberr, could you also update this in a jfrog-token module? We prefer to keep both of them in sync.

dimberr:
hi @matifali , I updated both jfrog-oauth and jfrog-token.
The use case is simple, we have multiple jfrog-cli configurations, every configuration has different server-id. 
In order to send request to specific configured server, jfrog cli has flag `--server-id`
We want to be able to specify server-id that is configured via this module

matifali (approved):

dimberr:
@matifali how can I consume that version from master? it still points to old version without the change

matifali:
@dimberr It's not available as a tag yet, and I am waiting on #154 to push a new tag. For testing, you can always use the latest version from `main` by removing `version` and changing `source` to

```tf
module "jfrog" {
  source = "git::https://github.com/coder/modules.git//jfrog-oauth?ref=main"
  # ...
  # all other options.
}
```

PR #160, opened 2024-02-14
bump version to v1.0.4

matifali:


mafredri (approved):
Why does this need to be done manually now? 🤔 (Or was it always manual?)

matifali:
it was always manual. The GitHub action I added is not working. 
I have fixed that in #155. Let's see if that triggers correctly on next tag push.

PR #161, opened 2024-02-15
Add slug to code server

TheZoker:
This PR:
- Adds a `slug` var to the code-server module
- Align the url structure of the vscode-web module with the code-server module

### DIFF
--- a/code-server/main.tf
+++ b/code-server/main.tf
@@ -32,6 +32,12 @@ variable "display_name" {
   default     = "code-server"
 }
 
+variable "slug" {
+  type        = string
+  description = "The slug for the code-server application."
+  default     = "code-server"
+}
+
 variable "settings" {
   type        = map(string)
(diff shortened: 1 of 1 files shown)


TheZoker:
@mafredri Can you please give me feedback, if the url structure should be the one from the `vscode-web` module or from the `code-server` module? Thanks!

mafredri:
> @mafredri Can you please give me feedback, if the url structure should be the one from the `vscode-web` module or from the `code-server` module? Thanks!

The two modules are very separate programs, so there isn't much reason to unify them. Adding the `urlencode` is probably a good idea, but if the behavior differs between `?folder=` and omitting it, then that behavior should be retained. If the behavior is the same between the two forms then the current one is fine (i.e. which folder is opened by default).

TheZoker:
Alright, I reverted the change

mafredri (approved):

Issue #92, opened 2023-10-16
How to verify module by coder


masterwendu:
Hi,

I created the 2 exoscale modules for the registry.
I set verified to `false` because I wasn´t sure how to verify.

Is is possible to verify a module?

Moreover you have a bug on the registry website.
In the modules overview our exoscale modules are not verified.
But in the detailed module view it is. It seems that you set a condition wrong there.

stirby:
I see what you're referring to and will look into it today.

matifali:
@stirby, now we have module versioning. Do you think having automated tests is enough to put a verified badge on a module?

masterwendu:
Hello @stirby, @matifali 

Do you have any updates on this issue?

stirby:
@matifali We can verify modules with tests, but we should update the readme to reflect this. We should also mention that the module creator is responsible for maintenance if that's the policy we're going with.

matifali:
@stirby I am ok with this. This can go in our CONTRIBUTING.md file.

matifali:
Hi @masterwendu, could you update the maintainer GitHub to your GitHub account instead of @Whizus

masterwendu:
> Hi @masterwendu, could you update the maintainer GitHub to your GitHub account instead of @WhizUs

Hi,
Actually we want to maintain this as a company. As we are allowed to work on it during work we want to have our Logo there.

PR #165, opened 2024-02-17
Add new `node-via-nvm` module

TheZoker:
This PR adds a [node](https://github.com/nodejs/node) installer module. It is powered by [NVM](https://github.com/nvm-sh/nvm). By default it installs the latest version of node and makes that the default. But the user can also provide a list of node versions, they like to install and also a custom default node version as well. 

I'm not 100% sure about the naming. Other possible names for the module:
- `nvm`
- `node`
- `nodejs` 

What you you prefer? 

Note: `node` installs the latest version of node.


TheZoker:
@mafredri @matifali Would appreciate, if you guys could have a look at this. Thanks!

node-via-nvm/run.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+NODE_VERSIONS=("${NODE_VERSIONS}")
```
mafredri, line 3: Parenthesis give the wrong impression that this will be an array. In theory, the `"` could also lead to variable expansion given the right kind of input, so let's change that to `'` instead of `"`, we're not dealing with a shell variable anyway.

```suggestion
NODE_VERSIONS='${NODE_VERSIONS}'
```

node-via-nvm/run.sh
```diff
@@ -0,0 +1,40 @@
...
+BOLD='\033[0;1m'
+CODE='\033[36;40;1m'
+RESET='\033[0m'
+
+printf "$${BOLD}Installing nvm!$${RESET}\n"
+
+export NVM_DIR="${INSTALL_PREFIX}/nvm"
```
mafredri, line 10: It's nice to assign tf vars to shell vars so that we always deal with them in the same way, I'd suggest the following change:

```suggestion
# top of file
INSTALL_PREFIX='${INSTALL_PREFIX}'

# here
export NVM_DIR="$${INSTALL_PREFIX}/nvm"
```

node-via-nvm/run.sh
```diff
@@ -0,0 +1,40 @@
...
+
+output=$(curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/${NVM_VERSION}/install.sh | bash)
```
mafredri, line 12:
```suggestion
# top
NVM_VERSION='${NVM_VERSION}'

# here
output=$(curl -o- "https://raw.githubusercontent.com/nvm-sh/nvm/$${NVM_VERSION}/install.sh" | bash)
```

mafredri: This will fail silently on network errors, see:

```sh
bash-5.2$ false | bash; echo $?
0
```

Here `bash` masks the exit code of curl.

We also don't capture stderr output here, so the `Failed to install` echo below will potentially be empty or not contain the actual failure. Maybe OK but could be confusing.

node-via-nvm/run.sh
```diff
@@ -0,0 +1,40 @@
...
+if [ $? -ne 0 ]; then
+  echo "Failed to install nvm: $output"
+  exit 1
+fi
+printf "🥳 nvm has been installed\n\n"
+
+# Set up nvm in the current shell session
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```
mafredri, line 20: Why is the source (`.`) escaped?

```suggestion
[ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
```

node-via-nvm/run.sh
```diff
@@ -0,0 +1,40 @@
...
+export NVM_DIR="${INSTALL_PREFIX}/nvm"
+
+output=$(curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/${NVM_VERSION}/install.sh | bash)
+if [ $? -ne 0 ]; then
+  echo "Failed to install nvm: $output"
+  exit 1
+fi
+printf "🥳 nvm has been installed\n\n"
+
+# Set up nvm in the current shell session
```
mafredri, line 19:
```suggestion
# Set up nvm for the rest of the script.
```

node-via-nvm/run.sh
```diff
@@ -0,0 +1,40 @@
...
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+
+# Install each node version...
+IFS=',' read -r -a VERSIONLIST <<< "$${NODE_VERSIONS}"
+for version in "$${VERSIONLIST[@]}"; do
+  if [ -z "$version" ]; then
+    continue
+  fi
+  printf "🛠️ Installing node version $${CODE}$version$${RESET}...\n"
+  output=$(nvm install "$version")
```
mafredri, line 29: Let's capture stderr too, that'll usually include the relevant error.

```suggestion
  output=$(nvm install "$version" 2>&1)
```

node-via-nvm/run.sh
```diff
@@ -0,0 +1,40 @@
...
+  if [ $? -ne 0 ]; then
+    echo "Failed to install version: $version: $output"
+    exit 1
+  fi
+done
+
+# Set default if provided
+if [ -n "${DEFAULT}" ]; then
+  printf "🛠️ Setting default node version $${CODE}$DEFAULT$${RESET}...\n"
```
mafredri, line 38: This is a good example of why I suggest:

```sh
# top
DEFAULT='${DEFAULT}'
```

Because if the above TF var assignment is made, then `$DEFAULT` will work here, right now it won't since it doesn't exist in the shell, only TF.

node-via-nvm/run.sh
```diff
@@ -0,0 +1,40 @@
...
+  output=$(nvm alias default $DEFAULT)
```
mafredri, line 39:
```suggestion
  output=$(nvm alias default $DEFAULT 2>&1)
```

We capture output but don't use it, looks like the `[ $? -ne 0 ]` check is missing here?

mafredri (commented):
Lots of comments, but a lot of them are the same suggestion repeated throughout. Hope it's not too overwhelming and thanks for creating this PR!

node-via-nvm/README.md
```diff
@@ -0,0 +1,58 @@
+---
+display_name: node-via-nvm
+description: Install node via nvm
+icon: ../.icons/node.svg
+maintainer_github: coder
```
matifali, line 5:
```suggestion
maintainer_github: TheZoker
```
Should we keep this module as official? We should keep as few modules as official as we can easily maintain.

node-via-nvm/run.sh
TheZoker: How can this be fixed in the code?

TheZoker:
Thanks for the review! I think fixed all issue but one (see comment)

nodejs/README.md
```diff
@@ -0,0 +1,58 @@
+---
+display_name: nodejs
+description: Install Node.js via nvm
+icon: ../.icons/node.svg
+maintainer_github: TheZoker
+verified: true
```
matifali, line 6: @mafredri @bpmct What is your opinion on accepting modules as verified?
There is a previous request in #92

matifali (changes requested):

nodejs/README.md
TheZoker: Ah I just copied that from another module, I can also remove it if requested

node-via-nvm/run.sh
mafredri: One option would be:

```sh
script="$(curl -o- "[url]")"
# handle error...
output="$(bash <<<"$script")"
# handle error...
```

mafredri: Updated curl with `-sS` so that we stay silent, but show error if there is one. That way the `2>&1` won't dirty the script output, but it'll still let us show an error if we failed and print `$script`.

TheZoker: Thanks!

mafredri: Saw your change, and you also need to handle the error after curl, i.e. `if [ $? ..` between script and output.

TheZoker:
Hmm very strange, I pushed changes to my branch and even rebased onto the latest `main`, but somehow github does not update this PR with the latest commit 😕 
https://github.com/TheZoker/coder-modules/commits/add-nvm-module/

@mafredri Any idea why this could be?

TheZoker:
@mafredri Alright, seems like the Github sync job has caught up. Can you check if the script is ok now? Thanks!

nodejs/run.sh
```diff
@@ -0,0 +1,50 @@
...
+CODE='\033[36;40;1m'
+RESET='\033[0m'
+
+printf "$${BOLD}Installing nvm!$${RESET}\n"
+
+export NVM_DIR="$${INSTALL_PREFIX}/nvm"
+
+script="$(curl -sS -o- "https://raw.githubusercontent.com/nvm-sh/nvm/$${NVM_VERSION}/install.sh" 2>&1)"
+if [ $? -ne 0 ]; then
+  echo "Failed to download nvm installation script."
```
mafredri, line 17:
```suggestion
  echo "Failed to download nvm installation script: $script"
```

nodejs/run.sh
```diff
@@ -0,0 +1,50 @@
...
+  printf "🛠️ Installing node version $${CODE}$version$${RESET}...\n"
+  output=$(nvm install "$version" 2>&1)
+  if [ $? -ne 0 ]; then
+    echo "Failed to install version: $version: $output"
+    exit 1
+  fi
+done
+
+# Set default if provided
+if [ -n "${DEFAULT}" ]; then
```
mafredri, line 47: This is using the TF var directly, one of the two forms (`$${foo}`, `$foo`) is necessary (unfortunately) for now:

```suggestion
if [ -n "$${DEFAULT}" ]; then
```

mafredri (commented):
Minor stuff, but this is mostly OK by me now. There's still the question of verified or not so I'll defer approval until that's been decided/taken care of.

nodejs/README.md
matifali: We originally published those modules and plan to maintain them so they are verified. We have yet to devise a criteria to mark community contributions as verified.

TheZoker: I see. So I would suggest to remove this line for now?

matifali: Yes, setting it ad `verified: false` until we decide on a criteria makes sense. We can change statues in a follow up PR.

TheZoker:
@mafredri Since we set `verified` to `false` for now, I think this issue should now be resolved, right?

TheZoker:
Thanks for the reviews and the merge 🎉

PR #170, opened 2024-02-23
Add terraform validation to linting

### LINKED ISSUE #143: run terraform validate as part of linting/formatting
@kylecarbs @stirby. I guess we already run a terraform apply for modules that have tests. So it might be unnecessary. 
Please correct me if I am wrong.

matifali:
This pull request adds terraform validation to the existing linting process. It includes a new script that runs terraform init and validate in each subdirectory containing a main.tf file. This ensures that the terraform code is valid and follows best practices.

Closes #143

### DIFF
--- a/terraform_validate.sh
+++ b/terraform_validate.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# Function to run terraform init and validate in a directory
+run_terraform() {
+  local dir="$1"
+  echo "Running terraform init and validate in $dir"
+  cd "$dir" || exit
+  terraform init
+  terraform validatecd 
+  cd - || exit
+}
+
+# Main script
+main() {
+  # Get the current directory
+  current_dir=$(pwd)
+
+  # Find all subdirectories containing a main.tf file
+  subdirs=$(find "$current_dir" -type f -name "main.tf" -exec dirname {} \;)
+
+  # Run terraform init and validate in each subdirectory
+  for dir in $subdirs; do
+    run_terraform "$dir"
+  done
+}
+
+# Run the main script
+main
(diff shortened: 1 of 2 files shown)


terraform_validate.sh
```diff
@@ -0,0 +1,28 @@
...
+  cd "$dir" || exit
+  terraform init
+  terraform validate
+  cd - || exit
+}
+
+# Main script
+main() {
+  # Get the current directory
+  current_dir=$(pwd)
```
mafredri, line 16: This will work from package JSON but not necessarily from shell, I'd recommend using the directory the script is in as a reference point.

terraform_validate.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# Function to run terraform init and validate in a directory
+run_terraform() {
+  local dir="$1"
+  echo "Running terraform init and validate in $dir"
+  cd "$dir" || exit
+  terraform init
+  terraform validate
+  cd - || exit
```
mafredri, line 10: Maybe we should just set -x on this script instead of handling failures manually.

mafredri:
> For some reason, I cannot get a nonzero exit code even if `terraform validate` shows errors.

I believe that would be fixed by `set -e` (or even better, `set -euo pipefail`).

terraform_validate.sh
```diff
@@ -0,0 +1,28 @@
...
+
+set -euo pipefail
+
+# Function to run terraform init and validate in a directory
+run_terraform() {
+    local dir="$1"
+    echo "Running terraform init and validate in $dir"
+    cd "$dir" || exit
+    terraform init -upgrade && terraform validate
+    cd - || exit
```
mafredri, line 11: We're already using `set -e`, no need for `&&` or `||`.
```suggestion
    pushd "$dir"
    terraform init -upgrade
    terraform validate
    popd
```

mafredri (approved):

PR #176, opened 2024-02-29
Add `app_name` parameter to `code-server`

TheZoker:


TheZoker:
An alternative would be to just use the `display_name` for the app name instead of introducing a new parameter

code-server/run.sh
```diff
@@ -47,4 +47,4 @@ fi
 
 echo "👷 Running code-server in the background..."
 echo "Check logs at ${LOG_PATH}!"
-$CODE_SERVER --auth none --port ${PORT} > ${LOG_PATH} 2>&1 &
+$CODE_SERVER --auth none --port ${PORT} --app-name ${APP_NAME} > ${LOG_PATH} 2>&1 &
```
code-asher, line 50: I am not sure if we sanitize variables but I wonder if we should add quotes in case the name contains spaces.  Hopefully no one puts quotes in their name, but maybe we also want to escape quotes that might be in the app name just in case or add sanitization if we do not already have it.

```suggestion
$CODE_SERVER --auth none --port ${PORT} --app-name "${APP_NAME}" > ${LOG_PATH} 2>&1 &
```

code-asher (commented):
Nice!  I think using `display_name` would make sense, we could always add a new variable later if we learn folks need to control it separately.

code-server/run.sh
TheZoker: Thanks for the suggestions! So I added the quotes, but for now did not escaped the input. As you said, we can later deal with this, if it is becomes an issue.

code-asher (approved):
Great, looks good to me, thank you!

TheZoker:
@matifali Thanks! Will there be a "release" soon?

Issue #181, opened 2024-03-06
docs: coder-server does not support offline mode


michaelbrewer:
[code-server](https://registry.coder.com/modules/code-server) docs mentions an offline mode, but this has not been implemented

```terraform
module "code-server" {
  source         = "registry.coder.com/modules/code-server/coder"
  version        = "1.0.7"
  agent_id       = coder_agent.main.id
  folder         = "/workspace/${local.folder_name}"
  install_prefix = "/workspace/.coder-server"
  offline = true
  settings = {
    "workbench.colorTheme" : "Visual Studio Dark"
    "workbench.startupEditor" : "readme"
  }
  extensions = [
    "ms-python.python",
    "hashicorp.terraform",
    "ms-azuretools.vscode-docker"
  ]
}
```

However there is no filed for this

```
Error: Unsupported argument
on main.tf line 219, in module "code-server":
  219:   offline = true
An argument named "offline" is not expected here.
```

---

Of coz having offline mode support by this module would be ideal

PR #184, opened 2024-03-12
feat(code-server): add code-server offline support

michaelbrewer:
## Changes

- Add "used_cached" which uses the cached copy of `code-server` if found, otherwise installs it from Github along with any extensions
- Add "offline" which uses the offline copy of `code-server`, and if not found returns an error
- Both modes applies any settings
- Both modes can not be used at the same time
- `offline` mode does not allow for extensions to be installed

## Checklist

- [x] Implement use_cached mode
- [x] Implement offline mode
- [ ] Add more notes on offline support and reference other code documentation
- [x] Add validation checks
- [x] Add terraform unit tests
- [ ] Add bash unit tests
- [ ] Update docs to reference guides on air-gapped installs

## Documentation examples

By default the module looks for code-server at `/tmp/code-server` but this can be changed with `install_prefix`.

Run an existing copy of code-server if found, otherwise download from GitHub:

```tf
module "code-server" {
  source     = "registry.coder.com/modules/code-server/coder"
  version    = "1.0.8"
  use_cached = true
  extensions = ["dracula-theme.theme-dracula", "ms-azuretools.vscode-docker"]
}
```

Just run code-server in the background, don't fetch it from GitHub:

```tf
module "code-server" {
  source   = "registry.coder.com/modules/code-server/coder"
  version  = "1.0.8"
  agent_id = coder_agent.example.id
  offline  = true
}
```

## Test cases

When offline is true and there is no copy of code-server is found 

```terraform
module "code-server" {
  source     = "registry.coder.com/modules/code-server/coder"
  version    = "1.0.8"
  offline    = true
}
```

Then expect runtime error

```
Failed to find a copy of code-server
```

---

When use_cached is true and there is no copy of code-server is found 

```terraform
module "code-server" {
  source     = "registry.coder.com/modules/code-server/coder"
  version    = "1.0.8"
  use_cached = true
  extensions = ["dracula-theme.theme-dracula", "ms-azuretools.vscode-docker"]
}
```

Then expect code-server to be installed

```
Installing code-server!
🥳 code-server has been installed in /home/local/.code-server
🧩 Installing extension dracula-theme.theme-dracula...
🧩 Installing extension ms-azuretools.vscode-docker...
👷 Running code-server in the background...
Check logs at /tmp/code-server.log!
```

---

When use_cached is true and offline is true

```terraform
module "code-server" {
  source     = "registry.coder.com/modules/code-server/coder"
  version    = "1.0.8"
  offline    = true
  use_cached = true
}
```

Then expect terraform validation error

```
Offline and Use Cached can not be used together
```

---

When offline is true and extensions are set

```terraform
module "code-server" {
  source     = "registry.coder.com/modules/code-server/coder"
  version    = "1.0.8"
  offline    = true
  extensions = ["dracula-theme.theme-dracula", "ms-azuretools.vscode-docker"]
}
```

Then expect terraform validation error

```
Offline mode does not allow extensions to be installed
```

---

When offline or use_cached is true and a copy of code-server is found then just run code-server in background

```
🥳 Found a copy of code-server
👷 Running code-server in the background...
Check logs at /tmp/code-server.log!
```

## Notes on air-gapped installations

Some cases for running `offline` installs of `code-server` with extensions, is for performance reasons, or because the egress controls on the workspace is tightly controlled or is completely air-gapped.

There is also cases where the coder terraform provisioner is also air-gapped,  for which you might have modules hosted on an [air-gapped terraform enterprise](https://developer.hashicorp.com/terraform/enterprise/registry/airgapped-providers), other methods are in the [coder documentation](https://coder.com/docs/v2/latest/templates/modules#offline-installations). It might also be useful to link to the [coder offline docs](https://coder.com/docs/v2/latest/install/offline)

related to #181


michaelbrewer:
@matifali  - this is a draft PR to add support for "offline" and maybe an option for download caching (so it only installed on first launch)

code-asher:
Is this ready for review or should I wait until it comes out of draft?

michaelbrewer:
> Is this ready for review or should I wait until it comes out of draft?

It's a draft, as I want to see if it makes sense to add an option to use cached installed too

michaelbrewer:
@code-asher it's ready to look at.

I did observe a weird behavior, when running from a cached version the "code-server" is available sooner, but the window is blank intialy. Also when it is open resizing the  window does not repaint.

code-server/README.md
```diff
@@ -79,3 +79,16 @@ module "code-server" {
   extensions = ["dracula-theme.theme-dracula", "ms-azuretools.vscode-docker"]
 }
 ```
+
+### Offline Mode
+
+Just run code-server in the background, don't fetch it from GitHub:
```
code-asher, line 85: Two thoughts here as I was trying out the module:

1. It could be helpful to indicate where to put code-server to make this work.  Here is a shot at some copy but it can probably be worded better:

```suggestion
Run an existing copy of code-server if found, otherwise download from GitHub. By default the module looks for code-server at `/tmp/code-server` but this can be changed with `install_prefix`.
```

2. Will users prefer if we throw an error when we cannot find code-server, given the `offline` name?  It could be surprising that `offline` could still pull from GitHub in some cases.  Or we could call this `use_existing` or `use_cached` instead or something like that, to avoid giving the impression it is fully offline.

code-asher (commented):
Nice!  I did not notice anything abnormal with the window when I ran with a cached copy.

code-server/README.md
michaelbrewer: I was wondering if "offline" should be strict like you use and should fail if not present. 

And I can add a "use_cache" flag for a speed boost on restarts.

matifali: Makes more sense to me.

code-server/README.md
```diff
@@ -79,3 +79,30 @@ module "code-server" {
...
+  use_cached = true
+  extensions = ["dracula-theme.theme-dracula", "ms-azuretools.vscode-docker"]
+}
+```
+
+Just run code-server in the background, don't fetch it from GitHub:
+
+```tf
+module "code-server" {
+  source   = "registry.coder.com/modules/code-server/coder"
```
matifali, line 103: `offline` mode will work great if `code-server` is prebuilt into the image. However, for offline deployments, they may not have access to registry.coder.com to fetch the module and can not use the feature unless they do something like [this](https://coder.com/docs/v2/latest/templates/modules#offline-installations).

michaelbrewer: @matifali  - true and for our use case we would use a [private module registry](https://developer.hashicorp.com/terraform/cloud-docs/registry/publish-modules) and for the running `code-server` instance, it would use the [code-marketplace](https://github.com/coder/code-marketplace).

But not your policy for the control plane running terraform could have egress, but the running workspace won't

michaelbrewer: @matifali  - Offline docs does cover alot of this, and companies might have flexibility to access `registry.terraform.io` but not github sh installs.

https://coder.com/docs/v2/latest/install/offline#offline-deployments

michaelbrewer: Terraform Enterprise also have good docs on how side load providers into an air gapped install of terraform enterprise. 

https://developer.hashicorp.com/terraform/enterprise/registry/airgapped-providers

michaelbrewer: Terraform Enterprise also have good docs on how side load providers into an air gapped install of terraform enterprise. 

https://developer.hashicorp.com/terraform/enterprise/registry/airgapped-providers

matifali: Thanks for the great PR and the explanations. 🎉

matifali (approved):

code-asher (approved):
Looks great!  I just have a question about a variable name in an error message.

code-server/main.tf
```diff
@@ -96,8 +108,22 @@ resource "coder_script" "code-server" {
...
     SETTINGS : replace(jsonencode(var.settings), "\"", "\\\""),
+    OFFLINE : var.offline,
+    USE_CACHED : var.use_cached,
   })
   run_on_start = true
+
+  lifecycle {
+    precondition {
+      condition     = !var.offline || length(var.extensions) == 0
+      error_message = "Offline mode does not allow extensions to be installed"
```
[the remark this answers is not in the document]
michaelbrewer: Makes sense, `offline` is the stricter form and may also bundle extensions in the image. But giving flexibility is also nice.

code-server/main.tf
```diff
@@ -96,8 +108,22 @@ resource "coder_script" "code-server" {
...
+    }
+
+    precondition {
+      condition     = !var.offline || !var.use_cached
+      error_message = "Offline and Use Cached can not be used together"
```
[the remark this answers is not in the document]
michaelbrewer: Terraform looks like this, so it is pretty clear. 

```bash
╷
│ Error: Resource precondition failed
│ 
│   on code-server/main.tf line 123, in resource "coder_script" "code-server":
│  123:       condition     = !var.offline || !var.use_cached
│     ├────────────────
│     │ var.offline is true
│     │ var.use_cached is true
│ 
│ Offline and Use Cached can not be used together
```

Issue #194, opened 2024-03-15
VS Code Web: Silence Curl


djarbz:
If these lines had the options `-Ss` then we would not see the extraneous curl output in the log.
https://github.com/coder/modules/blob/7e3743739edb0861fa6b569126066de4fc38af7f/vscode-web/run.sh#L22-L23

```
Installing Microsoft Visual Studio Code Server!
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

Checking if dotfiles repository already exists...
Found dotfiles repository at /home/coder/.config/coderv2/dotfiles

100  6279  100  6279    0     0  16266      0 --:--:-- --:--:-- --:--:-- 16266
```

michaelbrewer:
I have a related change that fixes this too

michaelbrewer:
@matifali  - this should be resolved now with the merge of #195

PR #195, opened 2024-03-15
feat(vscode-web): add support for settings

### LINKED ISSUE #196: VS Code Web: Add support for `settings`
Add support for pre-configuring settings with `vscode-web` like `code-server`

```terraform
module "vscode-web" {
  source     = "registry.coder.com/modules/vscode-web/coder"
  version    = "1.0.8"
  extensions = ["dracula-theme.theme-dracula"]
  settings = {
    "workbench.colorTheme" = "Dracula"
  }
  accept_license = true
}
```

michaelbrewer:
Add support for settings, like code-server

```tf
module "vscode-web" {
  source     = "registry.coder.com/modules/vscode-web/coder"
  version    = "1.0.8"
  extensions = ["dracula-theme.theme-dracula"]
  settings = {
    "workbench.colorTheme" = "Dracula"
  }
  accept_license = true
}
```

Currently saves to `~/.vscode-server/data/Machine/settings.json` as `~/.vscode-server/data/User/settings.json` does not seem to work.

related to #194
closes #196

### DIFF
--- a/vscode-web/README.md
+++ b/vscode-web/README.md
@@ -48,3 +48,20 @@ module "vscode-web" {
   accept_license = true
 }
 ```
+
+### Pre-configure Settings
+
+Configure VS Code's [settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settingsjson) file:
+
+```tf
+module "vscode-web" {
+  source     = "registry.coder.com/modules/vscode-web/coder"
+  version    = "1.0.8"
+  agent_id   = coder_agent.example.id
+  extensions = ["dracula-theme.theme-dracula"]
+  settings = {
+    "workbench.colorTheme" = "Dracula"
+  }
+  accept_license = true
+}
+```
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -93,7 +93,7 @@ variable "order" {
 
 variable "settings" {
   type        = map(string)
-  description = "A map of settings to apply to code-server."
+  description = "A map of settings to apply to VS Code web."
   default     = {}
 }
 
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -91,6 +91,12 @@ variable "order" {
   default     = null
 }
 
+variable "settings" {
+  type        = map(string)
+  description = "A map of settings to apply to code-server."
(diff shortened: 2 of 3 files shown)


vscode-web/main.tf
```diff
@@ -91,6 +91,12 @@ variable "order" {
   default     = null
 }
 
+variable "settings" {
+  type        = map(string)
+  description = "A map of settings to apply to code-server."
```
code-asher, line 96:
```suggestion
  description = "A map of settings to apply to VS Code web."
```

vscode-web/README.md
```diff
@@ -48,3 +48,20 @@ module "vscode-web" {
   accept_license = true
 }
 ```
+
+### Pre-configure Settings
+
+Configure VS Code's [settings.json](https://code.visualstudio.com/docs/getstarted/settings#_settingsjson) file:
```
code-asher, line 54: Should we add a warning that user settings cannot be customized because they are stored in the browser, so only machine settings will work?  We could also recommend `code-server` if they absolutely need to have user settings, or recommend using settings sync.

code-asher (commented):
Looks good!

vscode-web/README.md
michaelbrewer: @code-asher - maybe for another PR, but i think it would be useful to list out know issues and benefits the differentiate to the options. 

ie: 
code-server:
-  is more functionally complete 
- has less issues
- does not support the microsoft marketplace
- does not support  some of the microsoft extensions like copilt, liveshare
- does not require wildcard urls
- etc..

vscode-web:
- has issues with port forwarding
- settings issues
- requires wildcard / subdomain = true.
- does not support self-hosted extensions
- supports the microsoft extensions
- etc..

code-asher: Agreed!  This is a good list.  code-server has a list of differences from Codespaces (https://github.com/coder/code-server/blob/main/docs/FAQ.md#whats-the-difference-between-code-server-and-github-codespaces) but not the self-hosted version, so we should probably add something there as well.  Some of the differences there apply to the self-hosted version as well though.

code-asher (approved):

PR #205, opened 2024-03-24
feat(code-server): add extension_dir variable

### LINKED ISSUE #204: feat(code-server): Expose `--extensions-dir` to allow for caching of extensions when using a custom `install_prefix` not in `~/`
In a setup where `/workspace` is your persistent volume and `~/` is ephemeral
when i install `code-server` into `/workspace` any extensions will still be stored in  `~/.local/share/code-server/extensions` which is lost on restart

Example: 

```terraform
module "code-server" {
  source         = "registry.coder.com/modules/code-server/coder"
  version        = "1.0.10"
  folder         = "${local.workspace_home}/${local.folder_name}"
  install_prefix = "${local.workspace_home}/.code-server"
  use_cached     = true
  order          = 2
  slug           = "coder-server"
  settings = {
    "workbench.colorTheme" : "Visual Studio Dark"
    "workbench.startupEditor" : "readme"
  }
   extensions = [
     "ms-python.python",
     "hashicorp.terraform",
     "ms-azuretools.vscode-docker"
  ]
}
```

A solution could be to expose an option to set the `--extensions-dir` flag, then all of the extensions would be persisted between restarts

```terraform
module "code-server" {
  source         = "registry.coder.com/modules/code-server/coder"
  version        = "1.0.10"
  folder         = "${local.workspace_home}/${local.folder_name}"
  install_prefix = "${local.workspace_home}/.code-server"
  extensions_dir = "${local.workspace_home}/.code-server/extensions"
  use_cached     = true
  order          = 2
  slug           = "coder-server"
  settings = {
    "workbench.colorTheme" : "Visual Studio Dark"
    "workbench.startupEditor" : "readme"
  }
   extensions = [
     "ms-python.python",
     "hashicorp.terraform",
     "ms-azuretools.vscode-docker"
  ]
}
```
----

## Alternative solution

Alternative solution is to `install_prefix` and `use_cached` to make the assumption as to where 
the extension_dir should be.

```terraform
module "code-server" {
  source         = "registry.coder.com/modules/code-server/coder"
  version        = "1.0.10"
  folder         = "${local.workspace_home}/${local.folder_name}"
  install_prefix = "${local.workspace_home}/.code-server"
  use_cached     = true
  order          = 2
  slug           = "coder-server"
  settings = {
    "workbench.colorTheme" : "Visual Studio Dark"
    "workbench.startupEditor" : "readme"
  }
   extensions = [
     "ms-python.python",
     "hashicorp.terraform",
     "ms-azuretools.vscode-docker"
  ]
}
```

More info:
- https://coder.com/docs/code-server/latest/FAQ#where-are-extensions-stored

michaelbrewer:
```terraform
module "code-server" {
  source         = "registry.coder.com/modules/code-server/coder"
  version        = "1.0.10"
  folder         = "${local.workspace_home}/${local.folder_name}"
  install_prefix = "${local.workspace_home}/.code-server"
  extensions_dir = "${local.workspace_home}/.code-server/extensions"
  use_cached     = true
  order          = 2
  slug           = "coder-server"
  settings = {
    "workbench.colorTheme" : "Visual Studio Dark"
    "workbench.startupEditor" : "readme"
  }
   extensions = [
     "ms-python.python",
     "hashicorp.terraform",
     "ms-azuretools.vscode-docker"
  ]
}
````

closes #204

### CODE REVIEW

**Review by matifali — APPROVED** [review]

File: code-server/main.tf [hunk 1]
```diff
@@ -95,6 +95,12 @@ variable "use_cached" {
   default     = false
 }
 
+variable "extensions_dir" {
+  type        = string
+  description = "Override the directory to store extensions in. By default extensions are in ~/.local/share/code-server/extensions."
```
code-asher, line 100: Should we say `$XDG_DATA_HOME/code-server/extensions`?  That will probably almost always be `~/.local/share` but maybe in some cases it will be configured differently.

code-asher (approved):

code-server/main.tf
michaelbrewer: True, i would also mention that the following code does not support `XDG_DATA_HOME`

```bash
if [ ! -f ~/.local/share/code-server/User/settings.json ]; then
  echo "⚙️ Creating settings file..."
  mkdir -p ~/.local/share/code-server/User
  echo "${SETTINGS}" > ~/.local/share/code-server/User/settings.json
fi
```

code-asher: Oh right good catch.  Maybe we should just set `default` to `~/.local/share/code-server/extensions`?

michaelbrewer: We could, it _might_ be a breaking change for some, but making the default value `~/.local/share/code-server/extensions` is much more explicit.

code-asher: Good point.  Maybe we just update the description here for this PR and we can try figuring out what to do with this later.

code-server/run.sh
```diff
@@ -57,7 +63,7 @@ for extension in "$${EXTENSIONLIST[@]}"; do
     continue
   fi
   printf "🧩 Installing extension $${CODE}$extension$${RESET}...\n"
-  output=$($CODE_SERVER --install-extension "$extension")
+  output=$($CODE_SERVER "$EXTENSION_ARG" --install-extension "$extension")
```
mafredri, line 66: Not an issue in this PR, but the "output" in an error scenario will most likely be bad if we don't capture stderr as well.

```suggestion
  output=$($CODE_SERVER "$EXTENSION_ARG" --install-extension "$extension" 2>&1)
```

mafredri (approved):

code-server/run.sh
michaelbrewer: @mafredri - this actually hides the error. Current experience does look right

Screenshot 2024-04-04 at 12 38 10 AM

michaelbrewer:
Anything else to look at @code-asher  @mafredri  ?

code-server/run.sh
mafredri: @michaelbrewer that error message is different, though. Failed installing vs failed to install. Might you have removed `$output` from the message as well?

michaelbrewer: I was just doing negative testing to verify if I saw a failure message. 

When i add the redirection, no error is returned. So the change you suggested is not enough. 

I can look at alternatives if needed.

michaelbrewer: I am referring to the suggested change:

`output=$($CODE_SERVER "$EXTENSION_ARG" --install-extension "$extension" 2>&1)`

Which when there is an error, nothing is printed out.

Leaving the code as is works for me, i see errors when there is one, and otherwise it works as before.

mafredri: @michaelbrewer alright, right you are. Seems this is actually a bug in `code-server`. It doesn't exit with a non-zero exit code when this happens. As such the `if` below won't ever work either.

This needs to be fixed in code-server but for now, lets leave out the 2>&1 and add a comment here explaining it doesn't work as intended due to bug.

mafredri: https://github.com/coder/code-server/issues/6367

mafredri: This should actually be fixed since v4.20.1 of code-server. Which version were you testing @michaelbrewer ?

michaelbrewer: code-server --version returns: `4.22.1 760d1318e98945d05133c6121f99541ca7a39bf8 with Code 1.87.2`

```bash
CODE_SERVER=/home/local/.code-server/bin/code-server 
$CODE_SERVER --version
4.22.1 760d1318e98945d05133c6121f99541ca7a39bf8 with Code 1.87.2

EXTENSION_ARG=--extensions-dir=/home/local/.code-server/extenions/
extension=not-found

output=$($CODE_SERVER "$EXTENSION_ARG" --install-extension "$extension")
Extension 'not-found' not found.
Make sure you use the full extension ID, including the publisher, e.g.: ms-dotnettools.csharp
Failed Installing Extensions: not-found

echo $?
0
```

michaelbrewer: @mafredri  - upstream M$ code does exit 1, but when i run the latest code-server it always exits 0.

mafredri: @michaelbrewer thanks for testing and reporting. Nothing we can do/improve here for the time being. (Short of detecting the text "Failed Installing Extensions", which is far from ideal.)

michaelbrewer: Now the question is this a bug or feature. 😃   

From the end-users point of view. You get notified the extension failed to install. But the rest is fine. And you can open your workspace.

code-asher:
Nothing from me, I am OK on merging.

If the exit code works in Microsoft's server, maybe it was recently fixed.  I think we have a pending update, I will see if it resolves the issue today.

code-asher:
Figured out the exit code issue, next release will fix this (will be promoted to a full release early next week).

Issue #207, opened 2024-03-28
JetBrains-Gateway: coder_parameter order can not be set


michaelbrewer:
Jetbrains parameter will always be first on the list

PR #223, opened 2024-04-11
feat: add coder_parameter_order to all data.coder_parameter fields

michaelbrewer:
Add `coder_parameter_order` variable to modules that have a data coder_parameter

- [x] aws-region
- [x] azure-region
- [x] dotfiles
- [x] exoscale-instance-type
- [x] exoscale-zone
- [x] gcp-region
- [x] git-config (has 2 parameters)

## Examples

> NOTE: add `coder_parameter_order` for `git-config` could be increments by one, `gcp-region` order should be higher to start with. So intervals by 10 is always safe

```tf
module "git-config" {
  source                = "registry.coder.com/modules/git-config/coder"
  version               = "1.0.11"
  agent_id              = coder_agent.example.id
  allow_email_change    = true
  coder_parameter_order = 10
}

module "gcp-region" {
  source                = "registry.coder.com/modules/gcp-region/coder"
  version               = "1.0.11"
  regions               = ["us", "europe"]
  coder_parameter_order = 20
}

// etc..
```

related to #207

### DIFF
--- a/git-config/main.tf
+++ b/git-config/main.tf
@@ -26,6 +26,17 @@ variable "allow_email_change" {
   default     = false
 }
 
+variable "coder_parameter_user_email_order" {
+  type        = number
+  description = "The order determines the position of the 'user_email' template parameter in the UI/CLI presentation. The lowest order is shown first and parameters with equal order are sorted by name (ascending order)."
+  default     = null
+}
+
+variable "coder_parameter_username_order" {
+  type        = number
+  description = "The order determines the position of the 'username' template parameter in the UI/CLI presentation. The lowest order is shown first and parameters with equal order are sorted by name (ascending order)."
+  default     = null
+}
 
 data "coder_workspace" "me" {}
 
@@ -34,6 +45,7 @@ data "coder_parameter" "user_email" {
   name         = "user_email"
   type         = "string"
   default      = ""
+  order        = var.coder_parameter_user_email_order
   description  = "Git user.email to be used for commits. Leave empty to default to Coder user's email."
   display_name = "Git config user.email"
   mutable      = true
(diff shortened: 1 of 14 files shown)


matifali (approved):

git-config/main.tf
```diff
@@ -26,6 +26,17 @@ variable "allow_email_change" {
...
+  type        = number
+  description = "The order determines the position of the 'user_email' template parameter in the UI/CLI presentation. The lowest order is shown first and parameters with equal order are sorted by name (ascending order)."
+  default     = null
+}
+
+variable "coder_parameter_username_order" {
+  type        = number
+  description = "The order determines the position of the 'username' template parameter in the UI/CLI presentation. The lowest order is shown first and parameters with equal order are sorted by name (ascending order)."
+  default     = null
+}
```
mafredri, line 39: Would it be feasible to just use one `coder_parameter_order` for all parameters? If someone wants to change these individually, then there may be something wrong with the module or how it's being used.

Why? This makes modules updates more risky. If we ever add a third parameter, it could end up anywhere when the user updates.

Perhaps we can still apply order if it's given, e.g. `order = var.coder_parameter_order ? var.coder_parameter_order + 0 : null`, `+ 1`, etc.

An author can reserve as much ordered space for a module as they wish.

```tf
module "git-config" {
  source                = "registry.coder.com/modules/git-config/coder"
  version               = "1.0.11"
  agent_id              = coder_agent.example.id
  allow_email_change    = true
  coder_parameter_order = 10
}

module "gcp_region" {
  source                = "registry.coder.com/modules/gcp-region/coder"
  version               = "1.0.11"
  regions               = ["us", "europe"]
  coder_parameter_order = 20
}
```

To clarify this, maybe we could name it something like `coder_parameter_group_order` or `coder_parameter_order_weight`, I'm open to better ideas 😄.

michaelbrewer: True, i can keep it all the same field name.

michaelbrewer: @mafredri  - i have updated the code and tests

mafredri (approved):
Nice work, thanks for the PR!

PR #226, opened 2024-04-15
feat(jetbrains-gateway): Allow fetching latest version dynamically

### LINKED ISSUE #217: feat(Jetbrains-Gateway): Specify IDE Version or Latest
I would like to request an option to use a specific or `latest` version of the selected IDE.

For example:

- [x] Case 1: We specify `2024.1` to use the `2024.1` version of the IDE.
- [x] Case 2: We specify `latest` and upon startup of the workspace the latest version of the selected IDE will be discovered and utilized. If a new version is released, we simply would need to restart the workspace to link the new version.
- [x] Case 3: We leave the input blank and the version hardcoded into the module is used.

matifali:
This pull request adds the ability to fetch the JetBrains IDE's latest version dynamically. It adds a new variable to fetch the latest version. The `data.http.jetbrains_ide_versions` data source retrieves the latest version information.

TODO: Add tests (probably in a separate PR)

Closes #217

### DIFF
--- a/jetbrains-gateway/main.tf
+++ b/jetbrains-gateway/main.tf
@@ -6,6 +6,10 @@ terraform {
       source  = "coder/coder"
       version = ">= 0.17"
     }
+    http = {
+      source  = "hashicorp/http"
+      version = ">= 3.0"
+    }
   }
 }
 
@@ -44,6 +48,12 @@ variable "coder_parameter_order" {
   type        = number
   description = "The order determines the position of a template parameter in the UI/CLI presentation. The lowest order is shown first and parameters with equal order are sorted by name (ascending order)."
   default     = null
+}
+
+variable "latest" {
+  type        = bool
+  description = "Whether to fetch the latest version of the IDE."
+  default     = true
 }
 
 variable "jetbrains_ide_versions" {
@@ -120,6 +130,11 @@ variable "jetbrains_ides" {
   }
 }
 
+data "http" "jetbrains_ide_versions" {
+  for_each = var.latest ? toset(var.jetbrains_ides) : toset([])
+  url      = "https://data.services.jetbrains.com/products/releases?code=${each.key}&latest=true&type=release"
+}
+
 locals {
   jetbrains_ides = {
     "GO" = {
@@ -128,6 +143,7 @@ locals {
       identifier    = "GO",
       build_number  = var.jetbrains_ide_versions["GO"].build_number,
       download_link = "https://download.jetbrains.com/go/goland-${var.jetbrains_ide_versions["GO"].version}.tar.gz"
+      version       = var.jetbrains_ide_versions["GO"].version
     },
     "WS" = {
       icon          = "/icon/webstorm.svg",
@@ -135,6 +151,7 @@ locals {
       identifier    = "WS",
       build_number  = var.jetbrains_ide_versions["WS"].build_number,
       download_link = "https://download.jetbrains.com/webstorm/WebStorm-${var.jetbrains_ide_versions["WS"].version}.tar.gz"
+      version       = var.jetbrains_ide_versions["WS"].version
     },
     "IU" = {
       icon          = "/icon/intellij.svg",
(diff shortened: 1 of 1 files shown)


jetbrains-gateway/main.tf
```diff
@@ -120,6 +140,11 @@ variable "jetbrains_ides" {
   }
 }
 
+data "http" "jetbrains_ide_versions" {
+  for_each = var.latest ? toset(var.jetbrains_ides) : toset([])
```
matifali, line 144: `var.latest` condition ensures we do not call the HTTP provider if latest is not requested.

jetbrains-gateway/main.tf
```diff
@@ -128,57 +153,74 @@ locals {
...
       identifier    = "RD",
       build_number  = var.jetbrains_ide_versions["RD"].build_number,
       download_link = "https://download.jetbrains.com/rider/JetBrains.Rider-${var.jetbrains_ide_versions["RD"].version}.tar.gz"
+      version       = var.jetbrains_ide_versions["RD"].version
     }
   }
+
+  icon          = try(lookup(local.jetbrains_ides, data.coder_parameter.jetbrains_ide.value).icon, "/icon/gateway.svg")
+  json_data     = var.latest ? jsondecode(data.http.jetbrains_ide_versions[data.coder_parameter.jetbrains_ide.value].response_body) : {}
+  key           = var.latest ? keys(local.json_data)[0] : ""
```
matifali, line 218: The keys in the response are different from our IDE identifiers. For example, JetBrains now returns code `IIU` for IntelliJ Idea Ultimate instead of `IU`. So I am capturing the key dynamically from the response.

jetbrains-gateway/main.tf
```diff
@@ -193,9 +235,9 @@ data "coder_parameter" "jetbrains_ide" {
   dynamic "option" {
     for_each = var.jetbrains_ides
     content {
-      icon  = lookup(local.jetbrains_ides, option.value).icon
-      name  = lookup(local.jetbrains_ides, option.value).name
-      value = lookup(local.jetbrains_ides, option.value).identifier
+      icon  = local.jetbrains_ides[option.value].icon
+      name  = local.jetbrains_ides[option.value].name
+      value = option.value
```
matifali, line 240: This achieves the same purpose with a simpler syntax.

mafredri (approved):

mafredri:
Nice work! Do you think we'll need to proxy the latest lookups so that we can cache it and avoid hitting JetBrains API all the time?

jetbrains-gateway/main.tf
```diff
@@ -46,6 +50,22 @@ variable "coder_parameter_order" {
...
+  description = "Whether to fetch the latest version of the IDE."
+  default     = true
+}
+
+variable "channel" {
+  type        = string
+  description = "JetBrains IDE release channel. Valid values are release and eap."
+  default     = "release"
+  validation {
+    condition     = can(regex("^(?:release|eap)$", var.channel))
```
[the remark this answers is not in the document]
matifali: It's probably a remenent if when I was trying to achieve fetching custom versions. I will update.

matifali:
Yes. But I am unclear in how can we do achieve the caching for latest. Any suggestions?

jetbrains-gateway/main.tf
```diff
@@ -46,6 +50,22 @@ variable "coder_parameter_order" {
   default     = null
 }
 
+variable "latest" {
+  type        = bool
+  description = "Whether to fetch the latest version of the IDE."
+  default     = true
```
matifali, line 56:
```suggestion
  default     = false
```
@mafredri, should we default `latest` to `false` so that the existing behavior to use the hardcoded version is retained?

matifali (commented):
self review

jetbrains-gateway/main.tf
mafredri: It’s certainly less risky, so I’m all for it 👍

PR #231, opened 2024-04-19
feat(code-server): auto install extensions

michaelbrewer:
Add a new flag `auto_install_extensions` which auto installed the recommended extensions found in `.vscode/extensions.json`

```tf
module "code-server" {
 source                   = "registry.coder.com/modules/code-server/coder"
  version                 = "1.0.13"
  agent_id                = coder_agent.main.id
  folder                  = "/home/${local.username}/modules"
  install_prefix          = "/home/${local.username}/.code-server"
  extensions_dir          = "/home/${local.username}/.code-server/extensions"
  use_cached              = true
}
```

### DIFF
--- a/code-server/run.sh
+++ b/code-server/run.sh
@@ -70,4 +70,24 @@ for extension in "$${EXTENSIONLIST[@]}"; do
   fi
 done
 
+if [ "${AUTO_INSTALL_EXTENSIONS}" = true ]; then
+  WORKSPACE_DIR="$HOME/"
+  if [ -n "${FOLDER}" ]; then
+    WORKSPACE_DIR="${FOLDER}"
+  fi
+
+  if ! command -v jq > /dev/null; then
+    echo "jq is required to install extensions from a workspace file."
+    exit 0
+  fi
+
+  if [ -f "$WORKSPACE_DIR/.vscode/extensions.json" ]; then
+    printf "🧩 Installing extensions from %s/.vscode/extensions.json...\n" "$WORKSPACE_DIR"
+    extensions=$(jq -r '.recommendations[]' "$WORKSPACE_DIR"/.vscode/extensions.json)
+    for extension in $extensions; do
+      $CODE_SERVER "$EXTENSION_ARG" --install-extension "$extension"
+    done
+  fi
+fi
+
 run_code_server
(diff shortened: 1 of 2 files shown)


code-server/run.sh
```diff
@@ -70,4 +70,24 @@ for extension in "$${EXTENSIONLIST[@]}"; do
   fi
 done
 
+if [ "${AUTO_INSTALL_EXTENSIONS}" = true ]; then
+  WORKSPACE_DIR="$HOME/"
```
code-asher, line 74: With the `/` here we could get a double slash below.  I think generally this works fine, although maybe there are cases where it does not?

code-server/run.sh
```diff
@@ -70,4 +70,24 @@ for extension in "$${EXTENSIONLIST[@]}"; do
...
+  if [ -n "${FOLDER}" ]; then
+    WORKSPACE_DIR="${FOLDER}"
+  fi
+
+  if ! command -v jq > /dev/null; then
+    echo "jq is required to install extensions from a workspace file."
```
code-asher, line 80: Not sure if this is overkill, but we could the Node bundled with code-server. 

```
extensions=$("$INSTALL_PREFIX/lib/node" -e "require(\"$WORKSPACE_DIR/.vscode/extensions.json\").recommendations.forEach((r) => console.log(r))")
```

code-asher (approved):
Neat!

code-server/run.sh
michaelbrewer: cool. i will look into that.

code-server/run.sh
michaelbrewer: Interesting. I do like `jq`, but if we know the path to `node` and it should always be present, then it makes sense.

michaelbrewer:
@code-asher i am not 100% convinced about the flag `auto_install_extensions`,  if it makes sense i will keep it,

I will back port these features to the `vscode-web` module to people can easily switch between the 2 versions.

code-asher:
> i am not 100% convinced about the flag auto_install_extensions, if it makes sense i will keep it,

I think code-server/VS Code already offers to install the recommended extensions after you open a project, right?  But this can cut down on the perceived time it takes to get coding, so I feel it is a worthwhile addition.

michaelbrewer:
More of the naming of the flag.  

I like the intention of this, as it allows for the get repo to specify what it wants.

code-asher:
> More of the naming of the flag.

Ohh yeah I see, I think the name is good.

PR #235, opened 2024-04-25
feat(vscode-web): add offline, use_cached, extensions_dir and auto_install_extensions

michaelbrewer:
- [x] Pending the reveiw and acceptance of #231

### DIFF
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -97,6 +97,30 @@ variable "settings" {
   default     = {}
 }
 
+variable "offline" {
+  type        = bool
+  description = "Just run code-server in the background, don't fetch it from GitHub"
+  default     = false
+}
+
+variable "use_cached" {
+  type        = bool
+  description = "Uses cached copy code-server in the background, otherwise fetched it from GitHub"
+  default     = false
+}
+
+variable "extensions_dir" {
+  type        = string
+  description = "Override the directory to store extensions in."
+  default     = ""
+}
+
+variable "auto_install_extensions" {
+  type        = bool
+  description = "Automatically install recommended extensions when code-server starts."
+  default     = false
+}
+
 resource "coder_script" "vscode-web" {
   agent_id     = var.agent_id
   display_name = "VS Code Web"
@@ -109,8 +133,25 @@ resource "coder_script" "vscode-web" {
     TELEMETRY_LEVEL : var.telemetry_level,
     // This is necessary otherwise the quotes are stripped!
     SETTINGS : replace(jsonencode(var.settings), "\"", "\\\""),
+    OFFLINE : var.offline,
+    USE_CACHED : var.use_cached,
+    EXTENSIONS_DIR : var.extensions_dir,
+    FOLDER : var.folder,
+    AUTO_INSTALL_EXTENSIONS : var.auto_install_extensions,
   })
   run_on_start = true
+
+  lifecycle {
+    precondition {
+      condition     = !var.offline || length(var.extensions) == 0
+      error_message = "Offline mode does not allow extensions to be installed"
+    }
+
+    precondition {
+      condition     = !var.offline || !var.use_cached
+      error_message = "Offline and Use Cached can not be used together"
+    }
+  }
 }
 
 resource "coder_app" "vscode-web" {
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -99,7 +99,7 @@ variable "settings" {
 
 variable "offline" {
   type        = bool
-  description = "Just run code-server in the background, don't fetch it from GitHub"
+  description = "Just run VS Code Web in the background, don't fetch it from GitHub"
   default     = false
 }
 
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -149,7 +149,7 @@ resource "coder_script" "vscode-web" {
 
     precondition {
       condition     = !var.offline || !var.use_cached
-      error_message = "Offline and Use Cached can not be used together"
+      error_message = "offline and use_cached can not be used together"
     }
   }
 }
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -105,7 +105,7 @@ variable "offline" {
 
 variable "use_cached" {
   type        = bool
-  description = "Uses cached copy code-server in the background, otherwise fetched it from GitHub"
+  description = "Uses cached copy VS Code Web in the background, otherwise fetched it from GitHub"
   default     = false
 }
 
@@ -117,7 +117,7 @@ variable "extensions_dir" {
 
 variable "auto_install_extensions" {
   type        = bool
-  description = "Automatically install recommended extensions when code-server starts."
+  description = "Automatically install recommended extensions when VS Code Web starts."
   default     = false
 }
 
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -105,7 +105,7 @@ variable "offline" {
 
 variable "use_cached" {
   type        = bool
-  description = "Uses cached copy of VS Code Web in the background, otherwise fetches it from internet"
+  description = "Uses cached copy of VS Code Web in the background, otherwise fetches it from internet."
   default     = false
 }
 
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -99,7 +99,7 @@ variable "settings" {
 
 variable "offline" {
   type        = bool
-  description = "Just run VS Code Web in the background, don't fetch it from GitHub"
+  description = "Just run VS Code Web in the background, don't fetch it from the internet."
   default     = false
 }
 
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -105,7 +105,7 @@ variable "offline" {
 
 variable "use_cached" {
   type        = bool
-  description = "Uses cached copy VS Code Web in the background, otherwise fetched it from GitHub"
+  description = "Uses cached copy of VS Code Web in the background, otherwise fetches it from internet"
   default     = false
 }
 
--- a/vscode-web/main.tf
+++ b/vscode-web/main.tf
@@ -149,7 +149,7 @@ resource "coder_script" "vscode-web" {
 
     precondition {
       condition     = !var.offline || !var.use_cached
-      error_message = "offline and use_cached can not be used together"
+      error_message = "Offline and Use Cached can not be used together"
     }
   }
 }
--- a/vscode-web/run.sh
+++ b/vscode-web/run.sh
@@ -11,7 +11,7 @@ if [ -n "${EXTENSIONS_DIR}" ]; then
 fi
 
 run_vscode_web() {
-  echo "👷 Running $VSCODE_WEB serve-local $EXTENSIONS_ARG --port ${PORT} --host 127.0.0.1 --accept-server-license-terms --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
+  echo "👷 Running $VSCODE_WEB serve-local $EXTENSION_ARG --port ${PORT} --host 127.0.0.1 --accept-server-license-terms --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
   echo "Check logs at ${LOG_PATH}!"
   "$VSCODE_WEB" serve-local "$EXTENSION_ARG" --port "${PORT}" --host 127.0.0.1 --accept-server-license-terms --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
 }
--- a/vscode-web/run.sh
+++ b/vscode-web/run.sh
@@ -96,4 +96,4 @@ if [ "${AUTO_INSTALL_EXTENSIONS}" = true ]; then
   fi
 fi
 
-run_code_server
+run_vscode_web
--- a/vscode-web/run.sh
+++ b/vscode-web/run.sh
@@ -33,7 +33,7 @@ if [ -f "$VSCODE_SERVER" ]; then
 fi
 # Offline mode always expects a copy of vscode-server to be present
 if [ "${OFFLINE}" = true ]; then
-  echo "Failed to find a copy of code-server"
+  echo "Failed to find a copy of VS Code Web"
   exit 1
 fi
 
--- a/vscode-web/run.sh
+++ b/vscode-web/run.sh
@@ -2,7 +2,7 @@
 
 BOLD='\033[0;1m'
 EXTENSIONS=("${EXTENSIONS}")
-VSCODE_SERVER="${INSTALL_PREFIX}/bin/code-server"
+VSCODE_WEB="${INSTALL_PREFIX}/bin/code-server"
 
 # Set extension directory
 EXTENSION_ARG=""
--- a/vscode-web/run.sh
+++ b/vscode-web/run.sh
@@ -13,7 +13,7 @@ fi
 run_vscode_web() {
   echo "👷 Running $VSCODE_SERVER serve-local --port ${PORT} --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
   echo "Check logs at ${LOG_PATH}!"
-  "$VSCODE_SERVER" serve-local "$EXTENSION_ARG" --port "${PORT}" --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
+  "$VSCODE_WEB" serve-local "$EXTENSION_ARG" --port "${PORT}" --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
 }
 
 # Check if the settings file exists...
--- a/vscode-web/run.sh
+++ b/vscode-web/run.sh
@@ -60,7 +60,7 @@ if [ $? -ne 0 ]; then
   echo "Failed to install Microsoft Visual Studio Code Server: $output"
   exit 1
 fi
-printf "$${BOLD}Microsoft Visual Studio Code Server has been installed.\n"
+printf "$${BOLD}VS Code Web has been installed.\n"
 
 # Install each extension...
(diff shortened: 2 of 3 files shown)


vscode-web/run.sh
```diff
@@ -28,29 +62,38 @@ if [ $? -ne 0 ]; then
...
+  fi
+
+  WORKSPACE_DIR="$HOME"
+  if [ -n "${FOLDER}" ]; then
+    WORKSPACE_DIR="${FOLDER}"
+  fi
+
+  if [ -f "$WORKSPACE_DIR/.vscode/extensions.json" ]; then
+    printf "🧩 Installing extensions from %s/.vscode/extensions.json...\n" "$WORKSPACE_DIR"
+    extensions=$(jq -r '.recommendations[]' "$WORKSPACE_DIR"/.vscode/extensions.json)
```
matifali, line 92: I think we should add a check for `jq` and if not found skip installation with a message in logs.

michaelbrewer: Will look into that.

matifali (approved):

vscode-web/main.tf
```diff
@@ -97,6 +97,30 @@ variable "settings" {
   default     = {}
 }
 
+variable "offline" {
+  type        = bool
+  description = "Just run code-server in the background, don't fetch it from GitHub"
```
code-asher, line 102: We seem to have a few different terms (`VS Code Web`, `VS Code Server`, and `code-server`) in here.  I think maybe `VS Code Web` might make the most sense given the title of the module.

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
...
+# Check if the settings file exists...
+if [ ! -f ~/.vscode-server/data/Machine/settings.json ]; then
+  echo "⚙️ Creating settings file..."
+  mkdir -p ~/.vscode-server/data/Machine
+  echo "${SETTINGS}" > ~/.vscode-server/data/Machine/settings.json
+fi
+
+# Check if code-server is already installed for offline or cached mode
+if [ -f "$VSCODE_SERVER" ]; then
+  if [ "${OFFLINE}" = true ] || [ "${USE_CACHED}" = true ]; then
```
code-asher, line 28: Not a blocker for this PR, but I just realized we might not install extensions (either the ones in the config or the ones in `extensions.json`) if `USE_CACHED` is set, is that how we want it?

code-asher (commented):
Looks good to me, but I think we should use the same name in all the comments/output/functions (maybe `VS Code Web`).

vscode-web/run.sh
michaelbrewer: When cached. When you add a recommended extension, it would be installed or prompt you to install. But it would be installed in the cached persistent store. So stop / start will find it in the cache already.

michaelbrewer: So by design really.

vscode-web/main.tf
michaelbrewer: Ok. Will look.

vscode-web/run.sh
code-asher: I was thinking about a case where someone already has code-server/VS Code Web installed in the image and has the the cache enabled, and because it already exists the script exits early here and never installs the extensions below.

Or, suppose someone is using the cache then adds a new extension to the list.  It will never get installed automatically either.

michaelbrewer: @code-asher  - maybe it can run the installs again but just ignore errors when the same extension already exists.

code-asher: That seems reasonable!

michaelbrewer: (i assume that would be faster than a full install. :) )

vscode-web/main.tf
code-asher: Just FYI, I only put a comment on this line, but there are other references to `code-server` that we should also update (six, if I counted right, seven if we want to also update the `Microsoft Visual Studio Code Server` line).  Sorry if I was unclear!

vscode-web/main.tf
```diff
@@ -97,6 +97,30 @@ variable "settings" {
...
 
+variable "offline" {
+  type        = bool
+  description = "Just run VS Code Web in the background, don't fetch it from GitHub"
+  default     = false
+}
+
+variable "use_cached" {
+  type        = bool
+  description = "Uses cached copy VS Code Web in the background, otherwise fetched it from GitHub"
```
matifali, line 108:
```suggestion
  description = "Uses cached copy of VS Code Web in the background, otherwise fetched it from GitHub"
```

vscode-web/main.tf
```diff
@@ -109,8 +133,25 @@ resource "coder_script" "vscode-web" {
...
+
+  lifecycle {
+    precondition {
+      condition     = !var.offline || length(var.extensions) == 0
+      error_message = "Offline mode does not allow extensions to be installed"
+    }
+
+    precondition {
+      condition     = !var.offline || !var.use_cached
+      error_message = "Offline and Use Cached can not be used together"
```
matifali, line 152:
```suggestion
      error_message = "offline and use_cached can not be used together"
```
When using the module, these are input names exposed to the user.

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
...
+  echo "⚙️ Creating settings file..."
+  mkdir -p ~/.vscode-server/data/Machine
+  echo "${SETTINGS}" > ~/.vscode-server/data/Machine/settings.json
+fi
+
+# Check if vscode-server is already installed for offline or cached mode
+if [ -f "$VSCODE_SERVER" ]; then
+  if [ "${OFFLINE}" = true ] || [ "${USE_CACHED}" = true ]; then
+    echo "🥳 Found a copy of VS Code Web"
+    run_code_server
```
matifali, line 30:
```suggestion
    run_vscode_web
```

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
...
 EXTENSIONS=("${EXTENSIONS}")
+VSCODE_SERVER="${INSTALL_PREFIX}/bin/code-server"
+
+# Set extension directory
+EXTENSION_ARG=""
+if [ -n "${EXTENSIONS_DIR}" ]; then
+  EXTENSION_ARG="--extensions-dir=${EXTENSIONS_DIR}"
+fi
+
+run_code_server() {
```
matifali, line 13:
```suggestion
run_vscode_web() {
```

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
...
+if [ -f "$VSCODE_SERVER" ]; then
+  if [ "${OFFLINE}" = true ] || [ "${USE_CACHED}" = true ]; then
+    echo "🥳 Found a copy of VS Code Web"
+    run_code_server
+    exit 0
+  fi
+fi
+# Offline mode always expects a copy of vscode-server to be present
+if [ "${OFFLINE}" = true ]; then
+  echo "Failed to find a copy of code-server"
```
matifali, line 36:
```suggestion
  echo "Failed to find a copy of VS Code Web"
```

vscode-web/run.sh
```diff
@@ -28,29 +62,38 @@ if [ $? -ne 0 ]; then
 fi
 printf "$${BOLD}Microsoft Visual Studio Code Server has been installed.\n"
```
matifali, line 63:
```suggestion
printf "$${BOLD}VS Code Web has been installed.\n"
```

vscode-web/run.sh
```diff
@@ -28,29 +62,38 @@ if [ $? -ne 0 ]; then
...
+  WORKSPACE_DIR="$HOME"
+  if [ -n "${FOLDER}" ]; then
+    WORKSPACE_DIR="${FOLDER}"
+  fi
+
+  if [ -f "$WORKSPACE_DIR/.vscode/extensions.json" ]; then
+    printf "🧩 Installing extensions from %s/.vscode/extensions.json...\n" "$WORKSPACE_DIR"
+    extensions=$(jq -r '.recommendations[]' "$WORKSPACE_DIR"/.vscode/extensions.json)
+    for extension in $extensions; do
+      $VSCODE_SERVER "$EXTENSION_ARG" --install-extension "$extension" --force
```
matifali, line 94:
```suggestion
      $VSCODE_WEB "$EXTENSION_ARG" --install-extension "$extension" --force
```

vscode-web/run.sh
```diff
@@ -28,29 +62,38 @@ if [ $? -ne 0 ]; then
...
+    done
+  fi
 fi
 
-echo "👷 Running ${INSTALL_PREFIX}/bin/code-server serve-local --port ${PORT} --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
-echo "Check logs at ${LOG_PATH}!"
-"${INSTALL_PREFIX}/bin/code-server" serve-local --port "${PORT}" --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
+run_code_server
```
matifali, line 99:
```suggestion
run_vscode_web
```

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
...
+# Set extension directory
+EXTENSION_ARG=""
+if [ -n "${EXTENSIONS_DIR}" ]; then
+  EXTENSION_ARG="--extensions-dir=${EXTENSIONS_DIR}"
+fi
+
+run_code_server() {
+  echo "👷 Running $VSCODE_SERVER serve-local --port ${PORT} --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
+  echo "Check logs at ${LOG_PATH}!"
+  "$VSCODE_SERVER" serve-local "$EXTENSION_ARG" --port "${PORT}" --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
```
matifali, line 16:
```suggestion
  "$VSCODE_WEB" serve-local "$EXTENSION_ARG" --port "${PORT}" --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
```

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
...
+VSCODE_SERVER="${INSTALL_PREFIX}/bin/code-server"
+
+# Set extension directory
+EXTENSION_ARG=""
+if [ -n "${EXTENSIONS_DIR}" ]; then
+  EXTENSION_ARG="--extensions-dir=${EXTENSIONS_DIR}"
+fi
+
+run_code_server() {
+  echo "👷 Running $VSCODE_SERVER serve-local --port ${PORT} --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
```
matifali, line 14:
```suggestion
  echo "👷 Running $VSCODE_WEB serve-local --port ${PORT} --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
```

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
 
 BOLD='\033[0;1m'
 EXTENSIONS=("${EXTENSIONS}")
+VSCODE_SERVER="${INSTALL_PREFIX}/bin/code-server"
```
matifali, line 5:
```suggestion
VSCODE_WEB="${INSTALL_PREFIX}/bin/code-server"
```

matifali (commented):
I think I have suggested all the name changes, but I could have missed some.

michaelbrewer:
> I think I have suggested all the name changes, but I could have missed some.

some of this was there from before and some of it was exactly working code from the code-server implementation. I have accepted all your suggestions, but i will need to retest and fix CI errors.

matifali:
Thank you, @michaelbrewer ❤️ . You are doing a great job of making modules super helpful.

michaelbrewer:
> Thank you, @michaelbrewer ❤️ . You are doing a great job of making modules super helpful.
> 
> Let's wait for an approval from @code-asher and I will merge and cut a release.

Would be nice to have official templates is composed of modules. :)

matifali:
> Would be nice to have official templates is composed of modules. :)

I have also thought about this and created an [issue](https://github.com/coder/coder/issues/11784). But Coder bundles starter templates within the product and when we add modules that are fetched from registry, this does not work on restricted or air-gapped deployments.

michaelbrewer:
Scanning through the examples you can see cases of this:

- `examples/templates/aws-devcontainer/main.tf`   - (`https://get.docker.com` and `https://registry.coder.com/modules/code-server` etc..)
- `examples/templates/aws-linux/main.tf` - (`https://code-server.dev/install.sh`)
-  `examples/templates/docker/main.tf` - (`https://code-server.dev/install.sh`)
etc..

vscode-web/main.tf
```diff
@@ -97,6 +97,30 @@ variable "settings" {
   default     = {}
 }
 
+variable "offline" {
+  type        = bool
+  description = "Just run VS Code Web in the background, don't fetch it from GitHub"
```
code-asher, line 102: I forgot we fetch this from microsoft.com, not github.com.  I see also that we use the terminology "the internet" in the `use_cached` setting so I think we can use that here as well.

```suggestion
  description = "Just run VS Code Web in the background, don't fetch it from the internet."
```

vscode-web/main.tf
```diff
@@ -97,6 +97,30 @@ variable "settings" {
...
+  default     = false
+}
+
+variable "use_cached" {
+  type        = bool
+  description = "Uses cached copy of VS Code Web in the background, otherwise fetches it from internet"
```
code-asher, line 108: The others have a period, so I added one here, but let me know if this is wrong.

```suggestion
  description = "Uses cached copy of VS Code Web in the background, otherwise fetches it from internet."
```

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
...
+VSCODE_WEB="${INSTALL_PREFIX}/bin/code-server"
+
+# Set extension directory
+EXTENSION_ARG=""
+if [ -n "${EXTENSIONS_DIR}" ]; then
+  EXTENSION_ARG="--extensions-dir=${EXTENSIONS_DIR}"
+fi
+
+run_vscode_web() {
+  echo "👷 Running $VSCODE_WEB serve-local --port ${PORT} --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
```
code-asher, line 14: This is missing `$EXTENSIONS_ARG`.  In the future it might make sense to pass these args to a function that logs them and then calls the binary with them, so new arguments are not missed.  We also have `serve-local` twice.

```suggestion
  echo "👷 Running $VSCODE_WEB serve-local $EXTENSIONS_ARG --port ${PORT} --host 127.0.0.1 --accept-server-license-terms --without-connection-token --telemetry-level ${TELEMETRY_LEVEL} in the background..."
```

vscode-web/run.sh
```diff
@@ -2,6 +2,40 @@
...
+  echo "Check logs at ${LOG_PATH}!"
+  "$VSCODE_WEB" serve-local "$EXTENSION_ARG" --port "${PORT}" --host 127.0.0.1 --accept-server-license-terms serve-local --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
```
code-asher, line 16: This was already like this, but it seems we have `serve-local` twice.

```suggestion
  "$VSCODE_WEB" serve-local "$EXTENSION_ARG" --port "${PORT}" --host 127.0.0.1 --accept-server-license-terms --without-connection-token --telemetry-level "${TELEMETRY_LEVEL}" > "${LOG_PATH}" 2>&1 &
```

vscode-web/run.sh
```diff
@@ -37,20 +69,31 @@ for extension in "$${EXTENSIONLIST[@]}"; do
...
 
-# Check if the settings file exists...
-if [ ! -f ~/.vscode-server/data/Machine/settings.json ]; then
-  echo "⚙️ Creating settings file..."
-  mkdir -p ~/.vscode-server/data/Machine
-  echo "${SETTINGS}" > ~/.vscode-server/data/Machine/settings.json
+if [ "${AUTO_INSTALL_EXTENSIONS}" = true ]; then
+  if ! command -v jq > /dev/null; then
+    echo "jq is required to install extensions from a workspace file."
+    exit 0
```
code-asher, line 82: Just occurred to me, should this be considered a failure?

```suggestion
    exit 1
```

code-asher (approved):
Thank you @matifali for all finding all the spots to update the name!  Absolutely perfect.

I made a few more minor suggestions that I missed the first time around, nothing worth blocking over though so feel free to merge.

Thank you @michaelbrewer for putting this together!

vscode-web/run.sh
michaelbrewer: I am not sure. From a user perspective it would be nice to continue without an error and print a warning

michaelbrewer: I guess there is where monitoring works when errors happen in a large deployment.

code-asher: Oh if `exit 1` prevents the workspace from starting then yeah, better to continue.

matifali: it doesn't prevent the workspace from starting but does show a warning on the page that the script was not successful.

code-asher: Ahh gotcha yeah `exit 1` seems like the right move then.

matifali: I think it's fine. If we don't fine jw we should just skip installing extensions and continue with the rest of the script.

code-asher: I have mixed feelings about it, because enabling autoinstall without `jq` is a bug the template author has to fix (I think, right?), and not showing the error could make it so no one notices.

On the other hand, it does seem annoying as a user that I cannot launch VS Code just because someone forgot to install `jq`. 

stderr gets highlighted in the terminal, right?  Maybe we output to stderr and then keep going.

Also if we use `node` instead, we avoid the problem entirely.

michaelbrewer: I will try the node to in another PR.

PR #237, opened 2024-04-25
changed maintainer_github to coder, added partner_github: nataindata

timquinlan:
getting 500 error when template tries to pull module.  t

### DIFF
--- a/apache-airflow/README.md
+++ b/apache-airflow/README.md
@@ -2,8 +2,9 @@
 display_name: airflow
 description: A module that adds Apache Airflow in your Coder template
 icon: ../.icons/airflow.svg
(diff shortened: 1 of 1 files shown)


stirby (approved):

matifali:
@timquinlan The module was added after the https://github.com/coder/modules/releases/tag/v1.0.12 and could not be used. I just published https://github.com/coder/modules/releases/tag/v1.0.13.
You should be able to test it with `1.0.13` now.

Additionally, to test modules that still need to be tagged. You can use

```tf
module "example" {
  source = "git::https://github.com/coder/modules.git//<MODULE-NAME>?ref=main"
}
```

PR #241, opened 2024-04-26
feat: Add github-upload-public-key module

f0ssel:
Automates uploading Coder public key to Github so users don't have to.

First build:
Screenshot 2024-05-02 at 2 35 35 PM

Subsequent builds:
Screenshot 2024-05-02 at 2 36 02 PM

Github Auth not linked:
Screenshot 2024-05-02 at 2 36 54 PM

Closes https://github.com/coder/customers/issues/542


github-upload-public-key/run.sh
```diff
@@ -0,0 +1,93 @@
...
+if [ -z "$CODER_OWNER_SESSION_TOKEN" ]; then
+  echo "No coder owner session token specified!"
+  exit 1
+fi
+
+if [ -z "$GITHUB_TOKEN" ]; then
+  echo "No GITHUB_TOKEN in the workspace environment!"
+  echo "The user must be authenticated with Github before this script can be run."
+  exit 1
+fi
```
kylecarbs, line 22: Even better, you can use `coder external-auth access-token github` to automatically fetch the token for a provider! Then it'd refresh as well.

Emyrk: Yup, `GITHUB_TOKEN` env var method is only because the `gh` cli does not support a more dynamic approach. I've been pushing everyone in the community to not use when they do not have to.

kylecarbs: Yeah, we should remove this from dogfood.

kylecarbs: https://github.com/coder/coder/pull/13106

github-upload-public-key/README.md
```diff
@@ -3,13 +3,13 @@ display_name: Github Upload Public Key
 description: Automates uploading Coder public key to Github so users don't have to.
 icon: ../.icons/github.svg
 maintainer_github: f0ssel
```
matifali, line 5:
```suggestion
maintainer_github: coder
```
We can change it to as official coder module too.

f0ssel:
Working on adding some tests

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,108 @@
...
+
+echo "Fetching GitHub public SSH keys..."
+GITHUB_KEYS_RESPONSE=$(
+  curl -L -s \
+    -w "%%{http_code}" \
+    -H "Accept: application/vnd.github+json" \
+    -H "Authorization: Bearer $GITHUB_TOKEN" \
+    -H "X-GitHub-Api-Version: 2022-11-28" \
+    https://api.github.com/user/keys
+)
```
matifali, line 69: I suggest we generalize to accept custom GitHub URLs for GitHub enterprises customers. 
We can do that by accepting an additional input and default to github.com as fallback. 

Bonus if we can auto detect the GitHub URL from the the configured external auth. 

cc: @Emyrk

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,108 @@
...
+echo "Uploading Coder public SSH key to GitHub..."
+CODER_PUBLIC_KEY_NAME="$CODER_ACCESS_URL Workspaces"
+UPLOAD_RESPONSE=$(
+  curl -L -s \
+    -X POST \
+    -w "%%{http_code}" \
+    -H "Accept: application/vnd.github+json" \
+    -H "Authorization: Bearer $GITHUB_TOKEN" \
+    -H "X-GitHub-Api-Version: 2022-11-28" \
+    https://api.github.com/user/keys \
```
matifali, line 96: Same observation here to use a custom GitHub URL.

matifali (commented):
I suggest we generalize GitHub URL too by making it an input to the module or better deducing it from the configured GitHub external authentication.

kylecarbs (approved):
This is absolutely sick!

github-upload-public-key/main.tf
```diff
@@ -0,0 +1,55 @@
...
+  type        = string
+  description = "The access URL of the workspace."
+  default     = ""
+}
+
+variable "owner_session_token" {
+  type        = string
+  description = "The owner session token of the workspace."
+  default     = ""
+}
```
kylecarbs, line 40: You should be able to pass these as env vars when running the script instead, that way we don't have to have these extra vars.

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+
+set -e
+
+CODER_ACCESS_URL="${CODER_ACCESS_URL}"
+CODER_OWNER_SESSION_TOKEN="${CODER_OWNER_SESSION_TOKEN}"
+CODER_EXTERNAL_AUTH_ID="${CODER_EXTERNAL_AUTH_ID}"
+GITHUB_API_URL="${GITHUB_API_URL}"
+
+if [ -z "$CODER_ACCESS_URL" ]; then
+  echo "No coder access url specified!"
```
kylecarbs, line 11: For all of these I might put the env var it's looking for instead. Might reduce some confusion.

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+
+if [ -z "$GITHUB_API_URL" ]; then
+  echo "No GitHub API URL specified!"
+  exit 1
+fi
+
+echo "Fetching GitHub token..."
+GITHUB_TOKEN=$(coder external-auth access-token $CODER_EXTERNAL_AUTH_ID)
+if [ $? -ne 0 ]; then
+  echo "Failed to fetch GitHub token!"
```
kylecarbs, line 33: The output in this scenario will actually be the URL.

See: https://coder.com/docs/v2/latest/cli/external-auth_access-token#description

It'd be nice to output similarly to slackme in this scenario:

https://github.com/coder/modules/blob/main/slackme/slackme.sh#L64

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+  exit 1
+fi
+if [ -z "$GITHUB_TOKEN" ]; then
+  echo "No GitHub token found!"
+  exit 1
+fi
```
kylecarbs, line 39: I don't think this can happen...

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+echo "GitHub token found!"
```
kylecarbs, line 40: This might be helpful for debugging purposes, but I don't think this adds much for the user because of the messages below.

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+
+echo "Fetching Coder public SSH key..."
```
kylecarbs, line 42:
```suggestion
echo "Fetching public key from Coder..."
```

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+PUBLIC_KEY_BODY=$(sed \$d <<< "$PUBLIC_KEY_RESPONSE")
+
+if [ "$PUBLIC_KEY_RESPONSE_STATUS" -ne 200 ]; then
+  echo "Failed to fetch Coder public SSH key with status code $PUBLIC_KEY_RESPONSE_STATUS!"
+  echo "$PUBLIC_KEY_BODY"
+  exit 1
+fi
+
+PUBLIC_KEY=$(jq -r '.public_key' <<< "$PUBLIC_KEY_BODY")
+echo "Coder public SSH key found!"
```
kylecarbs, line 60: Is there a situation where it's not found? I might omit this line.

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+
+if [ -z "$PUBLIC_KEY" ]; then
+  echo "No Coder public SSH key found!"
+  exit 1
+fi
+
+echo "Fetching GitHub public SSH keys..."
```
kylecarbs, line 67:
```suggestion
echo "Fetching public keys from GitHub..."
```

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+if [ "$GITHUB_KEYS_RESPONSE_STATUS" -ne 200 ]; then
+  echo "Failed to fetch Coder public SSH key with status code $GITHUB_KEYS_RESPONSE_STATUS!"
+  echo "$GITHUB_KEYS_RESPONSE_BODY"
+  exit 1
+fi
+
+GITHUB_MATCH=$(jq -r --arg PUBLIC_KEY "$PUBLIC_KEY" '.[] | select(.key == $PUBLIC_KEY) | .key' <<< "$GITHUB_KEYS_RESPONSE_BODY")
+
+if [ "$PUBLIC_KEY" = "$GITHUB_MATCH" ]; then
+  echo "Coder public SSH key is already uploaded to GitHub!"
```
kylecarbs, line 88:
```suggestion
  echo "Your Coder public key is already on GitHub!"
```

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+  exit 0
+fi
+
+echo "Coder public SSH key not found in GitHub keys!"
+echo "Uploading Coder public SSH key to GitHub..."
```
kylecarbs, line 93:
```suggestion
echo "Your Coder public key is not in GitHub. Adding it now..."
```

github-upload-public-key/run.sh
```diff
@@ -0,0 +1,114 @@
...
+UPLOAD_RESPONSE_STATUS=$(tail -n1 <<< "$UPLOAD_RESPONSE")
+UPLOAD_RESPONSE_BODY=$(sed \$d <<< "$UPLOAD_RESPONSE")
+
+if [ "$UPLOAD_RESPONSE_STATUS" -ne 201 ]; then
+  echo "Failed to upload Coder public SSH key with status code $UPLOAD_RESPONSE_STATUS!"
+  echo "$UPLOAD_RESPONSE_BODY"
+  exit 1
+fi
+
+echo "Coder public SSH key uploaded to GitHub!"
```
kylecarbs, line 114:
```suggestion
echo "Your Coder public key has been added to GitHub!"
```

kylecarbs (approved):
Minor language things that I think could be helpful!

github-upload-public-key/main.tf
matifali: Yeah I can't think of. Use case where thet will be different from the provider provided values.

f0ssel: I was able to do this 👍

matifali (approved):

CONTRIBUTING.md
```diff
@@ -10,6 +10,8 @@ To create a new module, clone this repository and run:
 
 A suite of test-helpers exists to run `terraform apply` on modules with variables, and test script output against containers.
 
+The testing suite must be able to run docker containers with the `--network=host` flag, which typically requires running the tests on linux as this flag does not apply to Docker Desktop for MacOS and Windows. MacOS users can work around this by using something like [Orbstack](https://orbstack.dev/) instead of Docker Desktop.
```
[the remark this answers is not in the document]
f0ssel: good addition i meant to make, ty

f0ssel:
I realize there is a conflict between making tests pass and the script actually working in workspaces, I think I need to handle the terraform injection (I didn't know this was actually happening at first) as well as env vars for tests.

f0ssel:
Okay input variable issues are fixed, screenshots are updated, I think this is g2g but I'll wait until the eod to merge so yall can take a look at the last couple changes since review.

PR #252, opened 2024-05-25
fix(code-server): USE_CACHED should still install extensions

### LINKED ISSUE #251: fix(code-server): add extensions to be auto-installed even when a cached copy of coder is found
Currently when we find a cached copy of code-server we run it and exit early. But any requested extensions are not installed.

A solution would be to check if `install_prefix`/extensions or `extensions_dir` directories already exist before skipping ahead

michaelbrewer:
When USE_CACHED is true and no extensions have been installed Then we should still try to install them

close #251 

---

## Tests performed

- `bun test -t code-server`

## Scenario: use_cached and  extensions_dir is set, but no extensions installed

```tf
module "code-server" {
  source         = "./code_server"
  order          = 1
  folder         = "/home/${local.username}/modules"
  install_prefix = "/home/${local.username}/.code-server"
  extensions_dir = "/home/${local.username}/.code-server/extensions"
  use_cached = true
}
```

Given `USE_CACHED` is `true` 
AND a cached copy of code-server
AND we have a list of extensions to install at startup
Then we expect the extensions to be installed

Output when no extensions was installed before 

```text
⚙️ Creating settings file...
Installing code-server!
🥳 code-server has been installed in /home/local/.code-server
🧩 Installing extensions from /home/local/modules/.vscode/extensions.json...
Installing extensions...
Installing extension 'hashicorp.terraform'...
Extension 'hashicorp.terraform' v2.30.1 was successfully installed.
👷 Running code-server in the background...
Check logs at /tmp/code-server.log!
```

Output if extensions was already installed before

```text
🧩 Installing extensions from /home/local/modules/.vscode/extensions.json...
Installing extensions...
Extension 'hashicorp.terraform' v2.30.1 is already installed. Use '--force' option to update to latest version or provide '@' to install a specific version, for example: 'hashicorp.terraform@1.2.3'.
Installing extensions...
Extension 'esbenp.prettier-vscode' v10.4.0 is already installed. Use '--force' option to update to latest version or provide '@' to install a specific version, for example: 'esbenp.prettier-vscode@1.2.3'.
👷 Running code-server in the background...
Check logs at /tmp/code-server.log!
```

## Scenario: offline = true 

```tf
module "code-server" {
  source         = "./code_server"
  order          = 1
  folder         = "/home/${local.username}/modules"
  install_prefix = "/home/${local.username}/.code-server"
  extensions_dir = "/home/${local.username}/.code-server/extensions"
  offline = true
}
```

Expected output

```text
🥳 Found a copy of code-server
👷 Running code-server in the background...
Check logs at /tmp/code-server.log!
```


michaelbrewer:
@mafredri  something weird is happening with `git-config` tests? (`bun test -t git-config` works locally)

michaelbrewer:
@mafredri  - this PR #254 resolves the unit tests failing due to the fix in the newer coder/coder module

michaelbrewer:
@code-asher  - here is the PR for fixing cases where code-server has been reinstalled, but not any of the extensions.

code-server/run.sh
```diff
@@ -10,6 +10,9 @@ CODE_SERVER="${INSTALL_PREFIX}/bin/code-server"
 EXTENSION_ARG=""
 if [ -n "${EXTENSIONS_DIR}" ]; then
   EXTENSION_ARG="--extensions-dir=${EXTENSIONS_DIR}"
+  EXTENSIONS_DIR="${EXTENSIONS_DIR}"
```
code-asher, line 13: This could be skipped, I think.
```suggestion
```

code-server/run.sh
```diff
@@ -10,6 +10,9 @@ CODE_SERVER="${INSTALL_PREFIX}/bin/code-server"
...
+else
+  EXTENSIONS_DIR="${INSTALL_PREFIX}/extensions"
```
code-asher, line 15: This may not be the default extensions directory, right?  It should be something like `$XDG_DATA_HOME/code-server/extensions` (or `$HOME/.local` if `XDG_DATA_HOME` is not set).  At least, for Linux.  I am not sure if these modules can run in macOS and Windows.

We could always set `--extensions-dir` so we know where the extensions are, although I suppose that would not be a backwards compatible change.

Maybe we could just always install extensions regardless of whether the directory exists, since if an extension already exists code-server will be fine with that.  That would have the advantage of installing new extensions that are added to the template, too.

michaelbrewer: True. Let me adjust to handle defaults better.

michaelbrewer: Always installing is slower for restarting a workspace. So I would like to get some speed up on restart. 

I will see how much complexity it will add.

michaelbrewer: Maybe. Only cache extensions if there is an explicit extensions directory set. 

Otherwise always install (and maybe handle caches where existing extensions are there). 


I will need to document test cases as I go.

michaelbrewer:
@code-asher  a little simpler to only support installing extension when the the EXTENSION_DIR is set

michaelbrewer:
@code-asher  you are right :). Installing again when there are ones that already exist is really fast

code-server/run.sh
```diff
@@ -27,7 +27,11 @@ fi
 
 # Check if code-server is already installed for offline or cached mode
 if [ -f "$CODE_SERVER" ]; then
-  if [ "${OFFLINE}" = true ] || [ "${USE_CACHED}" = true ]; then
+  if [ "${USE_CACHED}" = true ] && [ -n "${EXTENSIONS_DIR}" ] && [ ! -d "${EXTENSIONS_DIR}" ]; then
```
code-asher, line 30: I think it would be nice to describe the behavior, maybe in the use_cached var description.  Something like this maybe:

> "Uses cached copy code-server in the background, otherwise fetched it from GitHub. Additionally, if extensions_dir is set, auto_install_extensions will only run if that directory does not already exist."

code-asher (commented):
Code looks good!

code-server/run.sh
code-asher: Oh I guess `extensions` is also involved.

> "Uses cached copy code-server in the background, otherwise fetched it from GitHub. Additionally, if extensions_dir is set, extensions will only be installed if that directory does not already exist."

michaelbrewer: @code-asher  - sorry i made a further simplication.

code-asher (approved):
Awesome!!

PR #259, opened 2024-06-09
feat(code-server): skip extension if installed

michaelbrewer:
To aid faster restarts of workspace with extensions already cached, this change will skip if an extension was already installed.

Before change, extensions would be try to be installed, but this operation can be slow, also extensions are not upgraded

Screenshot 2024-06-08 at 11 47 54 PM

After change with `use_cached_extensions` set, extensions that are already installed are skipped

```tf
module "code-server" {
  source                  = "registry.coder.com/modules/code-server/coder"
  version                 = "1.0.15"
  agent_id                = coder_agent.main.id
  order                   = 1
  folder                  = "/home/${local.username}/modules"
  install_prefix          = "/home/${local.username}/.code-server"
  use_cached              = true
  extensions              = [
    "dracula-theme.theme-dracula"
  ]
  use_cached_extensions   = true
  extensions_dir          = "/home/${local.username}/.code-server/extensions"
}
```

Screenshot 2024-06-10 at 10 34 22 PM

This change only impacts when `use_cached_extensions` and `extensions_dir` are set. 

--- 

Otherwise on restarts, `--force` is used to ensure existing extensions are upgraded.

Screenshot 2024-06-10 at 10 32 36 PM

### DIFF
--- a/code-server/main.tf
+++ b/code-server/main.tf
@@ -95,6 +95,12 @@ variable "use_cached" {
   default     = false
 }
 
+variable "use_cached_extensions" {
+  type        = bool
+  description = "Uses cached copy of extensions, otherwise do a forced upgrade"
+  default     = false
+}
+
 variable "extensions_dir" {
   type        = string
   description = "Override the directory to store extensions in."
@@ -122,6 +128,7 @@ resource "coder_script" "code-server" {
     SETTINGS : replace(jsonencode(var.settings), "\"", "\\\""),
     OFFLINE : var.offline,
     USE_CACHED : var.use_cached,
+    USE_CACHED_EXTENSIONS : var.use_cached_extensions,
     EXTENSIONS_DIR : var.extensions_dir,
     FOLDER : var.folder,
     AUTO_INSTALL_EXTENSIONS : var.auto_install_extensions,
--- a/code-server/run.sh
+++ b/code-server/run.sh
@@ -18,6 +18,21 @@ function run_code_server() {
   $CODE_SERVER "$EXTENSION_ARG" --auth none --port "${PORT}" --app-name "${APP_NAME}" > "${LOG_PATH}" 2>&1 &
 }
 
+function extension_installed() {
+  if [ -z "${EXTENSIONS_DIR}" ]; then
+    return 1
+  fi
+  EXTENSIONS_FILE="${EXTENSIONS_DIR}/extensions.json"
+  if [ ! -f "$EXTENSIONS_FILE" ]; then
+    return 1
+  fi
+  if grep -q "\"$1\"" "$EXTENSIONS_FILE"; then
+    echo "Extension $1 was found in $EXTENSIONS_FILE."
+    return 0
+  fi
+  return 1
+}
+
 # Check if the settings file exists...
 if [ ! -f ~/.local/share/code-server/User/settings.json ]; then
   echo "⚙️ Creating settings file..."
@@ -63,6 +78,9 @@ for extension in "$${EXTENSIONLIST[@]}"; do
   if [ -z "$extension" ]; then
     continue
   fi
+  if extension_installed "$extension"; then
+    continue
+  fi
   printf "🧩 Installing extension $${CODE}$extension$${RESET}...\n"
   output=$($CODE_SERVER "$EXTENSION_ARG" --install-extension "$extension")
   if [ $? -ne 0 ]; then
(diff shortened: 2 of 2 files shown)


michaelbrewer:
@code-asher  turns out locally it is fast, it is actually slower when doing an install if it already exists.

matifali:
How about using `--install-extension --force`? This should auto update the extensions to latest versions too. 

Skipping altogether will stuck the extensions to the initially installed version and code-server will then prompt to update them after the workspace starts.

michaelbrewer:
> How about using `--install-extension --force`? This should auto update the extensions to latest versions too.
> 
> Skipping altogether will stuck the extensions to the initially installed version and code-server will then prompt to update them after the workspace starts.

Currently, i have set a very short lifecycle for workspace, so there are more restarts, so i am looking at all things to speed up start time.

Our problem is slow starts of coder workspaces. I am trying to allow for caching of extensions and skipping over those already installed.

`--force` would make startups even slower, once you have `code-server` open, it may still have an update, but at least you can be coding already.

michaelbrewer:
(so far terraform init, terraform plan is also a very slow like 16 secs, so i am looking for time anywhere i can find it)

michaelbrewer:
@matifali if installing extensions was non-blocking (or at least upgrading them) then we get a win win situation :) 

See related issue for running some tasks in the background : https://github.com/coder/coder/issues/12448

code-asher:
Weird that it is so slow.  I suppose it must be the query to the marketplace that is taking so long?

Any idea if it is faster to install them all at once?  `code-server --install-extension $ext1 --install-extension $ex2` etc?  Maybe it will make a single query to the marketplace and be faster.

If we do proceed with this, should there be another flag to control it?  `CACHE_EXTENSIONS` or something?

michaelbrewer:
> Weird that it is so slow.  I suppose it must be the query to the marketplace that is taking so long?
> 
> 
> 
> Any idea if it is faster to install them all at once?  `code-server --install-extension $ext1 --install-extension $ex2` etc?  Maybe it will make a single query to the marketplace and be faster.
> 
> 
> 
> If we do proceed with this, should there be another flag to control it?  `CACHE_EXTENSIONS` or something?  And then add `--force` so by default it updates the version.

Oh. I did not know that could be done. Will add that for sure.

michaelbrewer:
Will look at the "CACHE_EXTENSIONS" option (false by default), then use that when skipping. And the "-force" would be the default

michaelbrewer:
@code-asher  have a look. @matifali  it seems like the tests are failing again

code-asher (approved):
Awesome thank you!

code-server/main.tf
```diff
@@ -95,6 +95,12 @@ variable "use_cached" {
   default     = false
 }
 
+variable "use_cached_extensions" {
+  type        = bool
+  description = "Uses cached copy of extensions, otherwise do a forced upgrade"
```
code-asher, line 100: Ah wait I just realized this only works if you set a manual extensions directory.  Should we say something like:

```
  description = "Uses cached copy of extensions, otherwise do a forced upgrade.  Only works if you set a custom extension directory."
```

But, perhaps we could try `code-server --list-extensions` instead?

code-asher (changes requested):

code-server/main.tf
michaelbrewer: Awesome, i see that works in both cases.

```shell
.code-server/bin/code-server --list-extensions
dracula-theme.theme-dracula
esbenp.prettier-vscode
golang.go
hashicorp.terraform
```

code-server/run.sh
```diff
@@ -79,6 +57,21 @@ if [ ! -f "$CODE_SERVER" ] || [ "${USE_CACHED}" != true ]; then
...
 fi
 
+# Get the list of installed extensions...
+LIST_EXTENSIONS=$($CODE_SERVER --list-extensions $EXTENSION_ARG)
+readarray -t EXTENSIONS_ARRAY <<< "$LIST_EXTENSIONS"
+function extension_installed() {
+  if [ "${USE_CACHED_EXTENSIONS}" != true ]; then
+    return 1
+  fi
+  for _extension in "$${EXTENSIONS_ARRAY[@]}"; do
```
code-asher, line 67: Should this be:

```suggestion
  for _extension in "${EXTENSIONS_ARRAY[@]}"; do
```

michaelbrewer: I have to escape the $ as this is a templated script.

code-server/run.sh
```diff
@@ -57,14 +57,33 @@ if [ ! -f "$CODE_SERVER" ] || [ "${USE_CACHED}" != true ]; then
...
+      echo "Extension $1 was already installed."
+      return 0
+    fi
+  done
+  return 1
+}
+
 # Install each extension...
 IFS=',' read -r -a EXTENSIONLIST <<< "$${EXTENSIONS}"
 for extension in "$${EXTENSIONLIST[@]}"; do
```
michaelbrewer, line 78: @code-asher - see example of why I need to escape the $

code-server/run.sh
code-asher: Ohhhhhhhhhhhhhhhh right that makes sense, duh!

code-asher (approved):
Awesome!

PR #262, opened 2024-06-26
feat: add module for Web RDP

Parkreiner:


test.ts
```diff
@@ -29,8 +29,10 @@ export const runContainer = async (
   return containerID.trim();
 };
 
-// executeScriptInContainer finds the only "coder_script"
-// resource in the given state and runs it in a container.
+/**
+ * Finds the only "coder_script" resource in the given state and runs it in a
+ * container.
+ */
```
Parkreiner, line 35: Updated the comment formatting so that these would be public, and would show up when someone hovers over the function definition

test.ts
```diff
@@ -76,27 +78,22 @@ export const execContainer = async (
...
-          attributes: {
-            [key: string]: any;
-          };
-        },
-      ];
-    },
-  ];
+  };
+
+  resources: [TerraformStateResource, ...TerraformStateResource[]];
```
Parkreiner, line 96: Redefined this so that `resources` is an array with at least element, rather than an array with exactly one element. This typo was causing some of the other tests to do stuff with the `any` type; I'll be making a separate PR to clean up more stuff

test.ts
```diff
@@ -131,12 +129,13 @@ export const findResourceInstance = <T extends "coder_script" | string>(
...
 
-// testRequiredVariables creates a test-case
-// for each variable provided and ensures that
-// the apply fails without it.
-export const testRequiredVariables = (
+/**
+ * Creates a test-case for each variable provided and ensures that the apply
+ * fails without it.
+ */
+export const testRequiredVariables = <TVars extends Record<string, string>>(
```
Parkreiner, line 136: Added a type parameter so that you can define what properties you expect to be valid for the function, so that you can more easily catch accidental typos

test.ts
BrunoQuaresma: TIL: `/**` are public comments and `//` are private 🤦

windows-rdp/main.test.ts
```diff
@@ -0,0 +1,130 @@
...
+} from "../test";
+
+type TestVariables = Readonly<{
+  agent_id: string;
+  resource_id: string;
+  admin_username?: string;
+  admin_password?: string;
+}>;
+
+function findWindowsRpdScript(state: TerraformState): string | null {
```
johnstcn, line 16: nit: typo: should be `findWindowsRdpScript`
alternatively: `findWindowsRDPScript`

windows-rdp/main.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    resource_id: "bar",
+  });
+
+  it("Has the PowerShell script install Devolutions Gateway", async () => {
+    const state = await runTerraformApply<TestVariables>(import.meta.dir, {
+      agent_id: "foo",
+      resource_id: "bar",
+    });
+
+    const lines = findWindowsRpdScript(state)
```
johnstcn, line 53: What if this function returns `null`?

windows-rdp/main.test.ts
```diff
@@ -0,0 +1,130 @@
...
+     * got to be a bit too much, though.
+     *
+     * Regex is a little bit more verbose and pedantic than normal. Want to
+     * have some basic safety nets for validating the structure of the form
+     * entries variable after the JS file has had values injected. Even with all
+     * the wildcard classes set to lazy mode, we want to make sure that they
+     * don't overshoot and grab too much content.
+     *
+     * Written and tested via Regex101
+     * @see {@link https://regex101.com/r/UMgQpv/2}
```
johnstcn, line 85: 👍

windows-rdp/main.tf
```diff
@@ -0,0 +1,177 @@
...
+  default   = "coderRDP!"
+  sensitive = true
+}
+
+resource "coder_script" "windows-rdp" {
+  agent_id     = var.agent_id
+  display_name = "windows-rdp"
+  icon         = "https://svgur.com/i/158F.svg" # TODO: add to Coder icons
+  script = <<EOF
+  function Set-AdminPassword {
```
johnstcn, line 38: suggestion, non-blocking: extract this to a terraform template file

https://developer.hashicorp.com/terraform/language/functions/templatefile

windows-rdp/main.tf
```diff
@@ -0,0 +1,177 @@
...
+  agent_id     = var.agent_id
+  display_name = "Local RDP"
+  slug         = "rdp-docs"
+  icon         = "https://raw.githubusercontent.com/matifali/logos/main/windows.svg"
+  url          = "https://coder.com/docs/v2/latest/ides/remote-desktops#rdp-desktop"
+  external     = true
+}
+
+# For some reason this is not rendering, commented out for now
+# resource "coder_metadata" "rdp_details" {
```
johnstcn, line 157: Is there an issue to follow-up on this?
I don't see any other module defining a `coder_metadata` resource.

johnstcn (approved):
LGTM 👍 smoke-tested

windows-rdp/main.test.ts
Parkreiner: Oh, good catch. The modules codebase doesn't have a super strict TypeScript compiler, so some of the types aren't as accurate as they should be, and some null/undefined values aren't being surfaced when they should be

I was already planning to tighten things up, but I for sure need to handle the null case here

windows-rdp/main.tf
Parkreiner: Oh, actually I forgot about this. This was a holdover from Ben's initial prototype
@bpmct Is it safe to delete all this?

BrunoQuaresma (approved):
The JS code looks good, but since I don't have extensive context on this, I'm unsure if I'm the right person to approve it. Also, how should we conduct QA for this PR?

Edit: I was able to partially QA it, and it seems fine. It would be helpful to have instructions in the description on how to QA this feature.

windows-rdp/main.tf
bpmct: Yeah, it's safe to delete. Personally, I would love to render it but I couldn't seem to get it working

matifali:
@Parkreiner you can merge the PR and then we can cut a release which should make it available on registry.coder.com

windows-rdp/powershell-installation-script.tftpl
```diff
@@ -0,0 +1,85 @@
...
+    New-ItemProperty -Path 'HKLM:\System\CurrentControlSet\Control\Terminal Server\WinStations\RDP-Tcp' -Name "UserAuthentication" -Value 0 -PropertyType DWORD -Force
+    New-ItemProperty -Path 'HKLM:\System\CurrentControlSet\Control\Terminal Server\WinStations\RDP-Tcp' -Name "SecurityLayer" -Value 1 -PropertyType DWORD -Force
+    # Enable RDP through Windows Firewall
+    Enable-NetFirewallRule -DisplayGroup "Remote Desktop"
+}
+
+function Install-DevolutionsGateway {
+# Define the module name and version
+$moduleName = "DevolutionsGateway"
+$moduleVersion = "2024.1.5"
```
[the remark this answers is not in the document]
Parkreiner: Yeah, just because the DOM selectors in the JavaScript code are so specific, it felt really important to pin Devolutions Gateway to a specific version, to avoid breaking changes as the Devolutions team refactors their own UI

It would be nice to expose it as a mutable value, but I feel like it might be asking for trouble, and could cause things to break unexpectedly

PR #266, opened 2024-07-03
docs(windows-rdp): make sure video renders correctly

matifali:
It now renders the video on GitHub but I am not sure if it will on registry.coder.com.

Any way to test it before merging?


Parkreiner:
I'm so sorry for not getting to this earlier, Atif. I'll review this first thing tomorrow morning

Parkreiner (commented):
Been poking around this, and I don't think there's a very clear, obvious solution, short of making fixes to the Coder registry's codebase.

The main problem with this PR is that if the markdown is ever put into a context that doesn't understand GitHub-specific syntax, this is all the user sees:

Screenshot 2024-07-10 at 10 08 47 AM

It's not really clear that this is meant to be a video, or that something happens if you click the icon. You can see this if you open the README locally, so I'm going to assume that the Coder registry would have the same problem.

The next thing I tried was embedding full HTML:
```html
<video controls="" muted="" preload="metadata" style="width: 100%; aspect-ratio: auto;">
  <source src="https://github.com/coder/modules/assets/28937484/fb5f4a55-7b69-4550-ab62-301e13a4be02" type="video/webm">

  <p>
    Your browser does not support GitHub's video format.
    <a href="https://github.com/coder/modules/assets/28937484/fb5f4a55-7b69-4550-ab62-301e13a4be02">
      Click here to try downloading the file.
    </a>
  </p>
</video>
```
This works locally (and I would assume it would work for the Registry, too), but GitHub strips it out completely:
Screenshot 2024-07-10 at 10 13 44 AM

<hr />

I think a decent compromise for the moment would be to:
1. Keep the GitHub-specific formatting
2. Update the image to be a screenshot of the video (I can take care of this later today)
3. Later on, update the Coder registry code so that it transforms the GitHub video markdown into an actual <video> element

The [current code for the markdown seems pretty easy to grasp](https://github.com/coder/registry.coder.com/blob/main/util/github.ts), since we're letting the `marked` library take care of the ASTs. I'm hoping an update actually won't be that bad. I'll try to squeeze that in later this week

matifali:
Thank you for the detailed review and for investigating all options. I will leave this PR open and you can directly push it with the most appropriate solution.

Parkreiner (approved):
Went ahead and made a new thumbnail. Decided to put it in a more local folder, so things have better co-location

PR #274, opened 2024-07-23
update windows-rdp to take share variable

### LINKED ISSUE #271: WebRDP module should support share level
I want to share my WebRDP with other users on a Coder deployment.

The Windows WebRDP module creates a [`coder_app`](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/app) for accessing the portal. That resource does not inherit the [`share`](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/app#share) attribute from a module variable, so it can't be assigned.

This module and all others that create a `coder_app` should allow the share level to be passed in.

bcpeinhardt:
Closes #271


Parkreiner (approved):
This looks like a straightforward change. I hadn't worked with the `share` variable before, but it looks perfectly in line with what we're doing with the other modules

windows-rdp/main.tf
```diff
@@ -9,6 +9,15 @@ terraform {
   }
 }
 
+variable "share" {
+  type    = string
+  default = "owner"
+  validation {
```
matifali, line 15: We can still add a `description` as they are rendered in registry.coder.com/modules variables tab.

Parkreiner: Would it be good to add that value for the `share` variables in the other modules, too?
Definitely on board with more documentation, but I don't think we're doing this anywhere just yet

bcpeinhardt: Cool cool 😎 I can make an issue for adding descriptions to these fields/generally formalizing documentation for modules and do it in a separate PR.

matifali: Thank you 😊

PR #270, opened 2024-07-16
fix url schema for webrdp example

stirby:
Tested the corrected URL on dogfood.

Was

```tf
module "windows_rdp" {
  source      = "registry.coder.com/coder/module/windows-rdp"
  ...
}
```

Now

```tf
module "windows_rdp" {
  source      = "registry.coder.com/modules/windows-rdp/coder"
  ...
}
```

PR #277, opened 2024-07-25
fix: typo in web RDP example

bpmct:


matifali (approved):
Thank you Ben. We should probably have a lint check to not have this typo again. I am seeing this [2nd](https://github.com/coder/modules/pull/270) time.

PR #281, opened 2024-08-14
chore: remove package-lock.json and update deps

matifali:
We use `bun` for package managemnt and looks like `package-lock.json` was added as an artifact of #236.

Additioanly updates deps to availble latest versions.


phorcys420 (approved):
LGTM

stirby (approved):

PR #286, opened 2024-08-29
feat(filebrowser): support subdomain = false

### LINKED ISSUE #285: Filebrowser module: support for subdomain = false
The Filebrowser can be configured to use a different baseURL than / with the command line parameter -b / --baseurl.

This can be used to allow using subdomain = false like in the terminal or code-server module.

Seppdo:
Resolves #285. 

Add variables `workspace_name`, `owner_name` and `resource_name` to be able to infer the directory the module is served at. Set this directory in the filebrowser app via the `filebrowser config set --baseurl` function.

Added an example to readme.

### DIFF
--- a/filebrowser/README.md
+++ b/filebrowser/README.md
@@ -44,3 +44,16 @@ module "filebrowser" {
   database_path = ".config/filebrowser.db"
 }
 ```
+
+### Serve from the same domain (no subdomain)
+
+Make sure to set both workspace_name and owner_name.
+
+```tf
+module "filebrowser" {
+  source         = "registry.coder.com/modules/filebrowser/coder"
+  agent_id       = coder_agent.main.id
+  workspace_name = data.coder_workspace.me.name
+  owner_name     = data.coder_workspace_owner.me.name
+}
+```
--- a/filebrowser/README.md
+++ b/filebrowser/README.md
@@ -47,13 +47,11 @@ module "filebrowser" {
 
 ### Serve from the same domain (no subdomain)
 
-Make sure to set both workspace_name and owner_name.
-
 ```tf
 module "filebrowser" {
-  source         = "registry.coder.com/modules/filebrowser/coder"
-  agent_id       = coder_agent.main.id
-  workspace_name = data.coder_workspace.me.name
-  owner_name     = data.coder_workspace_owner.me.name
+  source     = "registry.coder.com/modules/filebrowser/coder"
+  agent_id   = coder_agent.main.id
+  subdomain  = false
+  agent_name = "main"
 }
 ```
--- a/filebrowser/README.md
+++ b/filebrowser/README.md
@@ -13,9 +13,10 @@ A file browser for your workspace.
 
 ```tf
 module "filebrowser" {
-  source   = "registry.coder.com/modules/filebrowser/coder"
-  version  = "1.0.8"
-  agent_id = coder_agent.example.id
+  source     = "registry.coder.com/modules/filebrowser/coder"
+  version    = "1.0.8"
+  agent_id   = coder_agent.example.id
+  agent_name = "main"
 }
 ```
 
@@ -27,10 +28,11 @@ module "filebrowser" {
 
 ```tf
 module "filebrowser" {
-  source   = "registry.coder.com/modules/filebrowser/coder"
-  version  = "1.0.8"
-  agent_id = coder_agent.example.id
-  folder   = "/home/coder/project"
+  source     = "registry.coder.com/modules/filebrowser/coder"
+  version    = "1.0.8"
+  agent_id   = coder_agent.example.id
+  agent_name = "main"
+  folder     = "/home/coder/project"
 }
 ```
 
@@ -41,6 +43,7 @@ module "filebrowser" {
   source        = "registry.coder.com/modules/filebrowser/coder"
   version       = "1.0.8"
   agent_id      = coder_agent.example.id
+  agent_name    = "main"
   database_path = ".config/filebrowser.db"
 }
 ```
@@ -50,8 +53,8 @@ module "filebrowser" {
 ```tf
 module "filebrowser" {
   source     = "registry.coder.com/modules/filebrowser/coder"
-  agent_id   = coder_agent.main.id
-  subdomain  = false
+  agent_id   = coder_agent.example.id
   agent_name = "main"
+  subdomain  = false
 }
 ```
--- a/filebrowser/main.tf
+++ b/filebrowser/main.tf
@@ -14,10 +14,8 @@ variable "agent_id" {
   description = "The ID of a Coder agent."
 }
 
-variable "workspace_name" {
-  type = string
-  default = ""
-  description = "Set this and owner_name to serve filebrowser from subdirectory."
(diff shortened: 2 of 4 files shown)


filebrowser/main.tf
```diff
@@ -67,7 +85,10 @@ resource "coder_script" "filebrowser" {
     PORT : var.port,
     FOLDER : var.folder,
     LOG_PATH : var.log_path,
-    DB_PATH : var.database_path
+    DB_PATH : var.database_path,
+    WORKSPACE_NAME : var.workspace_name,
+    OWNER_NAME : var.owner_name,
+    RESOURCE_NAME : var.resource_name
```
matifali, line 91:
```suggestion
    AGENT_NAME : var.agent_name
    SUBDOMAIN : var.subdomain
```

filebrowser/main.tf
```diff
@@ -78,7 +99,7 @@ resource "coder_app" "filebrowser" {
   display_name = "File Browser"
   url          = "http://localhost:${var.port}"
   icon         = "https://raw.githubusercontent.com/filebrowser/logo/master/icon_raw.svg"
-  subdomain    = true
+  subdomain    = var.owner_name == "" && var.workspace_name == ""
```
matifali, line 102:
```suggestion
  subdomain    = var.subdomain
```

filebrowser/run.sh
```diff
@@ -17,6 +17,13 @@ if [ "${DB_PATH}" != "filebrowser.db" ]; then
   DB_FLAG=" -d ${DB_PATH}"
 fi
 
+# set baseurl if subdomain = false (owner_name and workspace_name is set), else reset to "" for use with subdomain = true
+if [ "${OWNER_NAME}" != "" ] && [ "${WORKSPACE_NAME}" != "" ]; then
```
matifali, line 21:
```suggestion
# set baseurl if subdomain = false
if [ "${SUBDOMAIN}" == "false" ]; then
```

filebrowser/main.tf
```diff
@@ -14,6 +14,24 @@ variable "agent_id" {
...
+  type = string
+  default = ""
+  description = "Set this and workspace_name to serve filebrowser from subdirectory."
+}
+
+variable "resource_name" {
+  type = string
+  default = "main"
+  description = "The name of the main deployment. (Used to build the subdirectory of the module.)"
+}
```
matifali, line 33: Lets call it `agent_name` we 
```suggestion
variable "agent_name" {
  type = string
  default = ""
  description = "The name of the main deployment. (Used to build the subpath for coder_app.)"
}
```

filebrowser/main.tf
```diff
@@ -14,6 +14,24 @@ variable "agent_id" {
   description = "The ID of a Coder agent."
 }
 
+variable "workspace_name" {
+  type = string
+  default = ""
+  description = "Set this and owner_name to serve filebrowser from subdirectory."
+}
+
```
matifali, line 22: This can be fetched from https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/workspace
```suggestion
```

filebrowser/main.tf
```diff
@@ -14,6 +14,24 @@ variable "agent_id" {
...
+variable "owner_name" {
+  type = string
+  default = ""
+  description = "Set this and workspace_name to serve filebrowser from subdirectory."
+}
+
```
matifali, line 28: And this from https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/workspace_owner
```suggestion
```

matifali (changes requested):
Thank you @Seppdo for the contribution. 
I have requested a few changes to make it more optimized.

filebrowser/README.md
```diff
@@ -44,3 +44,16 @@ module "filebrowser" {
...
+Make sure to set both workspace_name and owner_name.
+
+```tf
+module "filebrowser" {
+  source         = "registry.coder.com/modules/filebrowser/coder"
+  agent_id       = coder_agent.main.id
+  workspace_name = data.coder_workspace.me.name
+  owner_name     = data.coder_workspace_owner.me.name
+}
+```
```
matifali, line 59:
```suggestion
### Serve from the same domain (no subdomain)

Make sure to set both workspace_name and owner_name.

```tf
module "filebrowser" {
  source         = "registry.coder.com/modules/filebrowser/coder"
  workspace_name = data.coder_workspace.me.name
  owner_name     = data.coder_workspace_owner.me.name
}
```
```
Also update the example accordingly.

File: filebrowser/main.tf [hunk 8]
```diff
@@ -67,7 +85,10 @@ resource "coder_script" "filebrowser" {
     PORT : var.port,
     FOLDER : var.folder,
     LOG_PATH : var.log_path,
-    DB_PATH : var.database_path
+    DB_PATH : var.database_path,
+    WORKSPACE_NAME : var.workspace_name,
+    OWNER_NAME : var.owner_name,
```
matifali, line 90:
```suggestion
    WORKSPACE_NAME : data.coder_workspace.me.name,
    OWNER_NAME : data.coder_workspace_owner.me.name,
```

matifali (changes requested):
You may have to adjust a few places.

matifali:
You can also check my review for https://github.com/coder/modules/pull/288 and do a similar approach. In this case you can discard the requested changes.

filebrowser/main.tf
```diff
@@ -78,7 +94,7 @@ resource "coder_app" "filebrowser" {
   display_name = "File Browser"
   url          = "http://localhost:${var.port}"
```
matifali, line 95: Don't we need to set the URL to a correctly formatted URL with a subpath here?

Seppdo: The Filebrowser app will set the given basepath if called with the root path (/). So there is no need to set the subpath beforehand.

Seppdo: The filebrowser app automatically redirects to the given basepath when called with the root path (/). So there is no need to change it here.
Also this is the correct url if called with subdomain = true.

Seppdo: As I understand, the subpath is stripped by the proxy, so the filebrowser app just sees the path behind the SERVER_BASE_PATH. So there is no need to change it.

filebrowser/main.tf
```diff
@@ -20,7 +20,7 @@ data "coder_workspace_owner" "me" {}
 
 variable "agent_name" {
   type = string
-  default = ""
+  default = "main"
```
matifali, line 23:
```suggestion

```
let us remove default to make `agent_name` a required input. What do you think?

Seppdo: Since a suggestion for "main" is given in the readme. This can be an required input.

Seppdo:
Since agent_name is a required parameter, it must be set even in cases where it is not actually needed (e.g. every case with subdomain=true)

Seppdo:
I don't know how to fix the error in the pretty check
`error: script "fmt:ci" exited with code 3`

matifali:
@Seppdo can you run `bun i && bun fmt`?

matifali (approved):

PR #299, opened 2024-09-24
fix(filebrowser): only require agent_name when not on subdomain

matifali:
Ensure that agent_name defaults to an empty string, and introduce
validation to require it when subdomain is true. This aligns with
expected behavior and prevents misconfiguration.


phorcys420:
hey @matifali, couldn't we just grab it somehow from the agent id?

Parkreiner (approved):
These look like really straightforward changes. Approving

PR #301, opened 2024-09-24
chore: move update-version to ci

### LINKED ISSUE #229: registry.coder.com shows wrong version
It looks like what is displayed on the website is the main branch, but the modules only can be accessed by a version tag.  This might be an artifact of #157 though.  This can cause confusion as the published documentation is not always correct.

matifali:
Remove the automated update workflow in favor of a check.
This only lets us bump by a patch version for now, which is fine as we will have separate versions for each module eventually. See #157

This should also fix #229


matifali:
@mafredri I could use your help in fixing the CI. the script works fine locally.

.github/workflows/ci.yaml
```diff
@@ -37,4 +39,14 @@ jobs:
       - name: typos-action
         uses: crate-ci/typos@v1.17.2
       - name: Lint
-        run: bun lint
+        run: |
+          bun lint
+          # check for version changes
+          ./update-version.sh
+          if [[ $(git status --porcelain) ]]; then
```
mafredri, line 46:
```suggestion
          if [[ -n "$(git status --porcelain)" ]]; then
```

update-version.sh
```diff
@@ -25,5 +37,19 @@ for dir in "${changed_dirs[@]}"; do
...
+    if ! git diff --quiet "$dir/README.md"; then
+      echo "Bumping version in $dir/README.md"
+    fi
   fi
 done
+
+# Check if any changes were made in README.md files
+if [[ -n "$(git status --porcelain -- '**/README.md')" ]]; then
+  echo "Version mismatch detected. Please commit the updated README.md files."
+  git diff '**/README.md'  # Show what is different in the README.md files
```
mafredri, line 51:
```suggestion
  git diff -- '**/README.md'  # Show what is different in the README.md files
```

mafredri (approved):
I think it's just the if condition in yaml that's failing. The other changes are optional.

matifali:
Thank you @mafredri. 
@Parkreiner, can we point our registry to the latest available tag instead of `main` after this PR, people will start seeing 1.0.19, which has not been pushed yet.

PR #308, opened 2024-09-28
feat(git-clone): add git@ URLs support

djarbz:
Add Github and Gitlab when using git@ URLs.


matifali:
Hi @djarbz, do you plan to come back to this? I see we have linting issues.

djarbz:
@matifali thanks for the reminder.
Should be G2G now.

matifali:
@michaelbrewer Could you review, too, as you are the original author? Thanks.

Issue #313, opened 2024-10-07
Subpath on Jupyterlab module


framctr:
Hi, I'm able to use Jupyter on subpath in the following way:

```hcl
data "coder_workspace" "me" {}
data "coder_workspace_owner" "me" {}

resource "coder_agent" "main" {
  os   = "linux"
  arch = "amd64"

  metadata {
    display_name = "CPU Usage"
    key          = "0_cpu_usage"
    script       = "coder stat cpu"
    interval     = 10
    timeout      = 1
  }

  metadata {
    display_name = "RAM Usage"
    key          = "1_ram_usage"
    script       = "coder stat mem"
    interval     = 10
    timeout      = 1
  }

  metadata {
    display_name = "Home Disk"
    key          = "3_home_disk"
    script       = "coder stat disk --path $${HOME}"
    interval     = 60
    timeout      = 1
  }
}

resource "coder_script" "jupyterlab" {
  agent_id     = coder_agent.main.id
  display_name = "Jupyter"
  icon         = "/icon/jupyter.svg"
  script = templatefile("${path.module}/run.sh",
    {
      LOG_PATH        = "/tmp/jupyterlab.log"
      PORT            = 8888
      WORKSPACE_OWNER = data.coder_workspace_owner.me.name
      WORKSPACE_NAME  = data.coder_workspace.me.name
    }
  )
  run_on_start = true
}

resource "coder_app" "jupyter" {
  agent_id     = coder_agent.main.id
  display_name = "JupyterLab"
  icon         = "/icon/jupyter.svg"
  url          = "http://localhost:8888/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyter"
  slug         = "jupyter" # keep in sync with end of URL
  subdomain    = false
  share        = "owner"

  healthcheck {
    url       = "http://localhost:8888/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyter"
    interval  = 6
    threshold = 10
  }
}
```

And in the `run.sh` I pass also `--ServerApp.base_url=/@"${WORKSPACE_OWNER}"/"${WORKSPACE_NAME}"/apps/jupyter` to the `jupyter-lab` command.

Do you think this change could be added to the jupyter module?

I see from issue coder/registry.coder.com#31 that you would like to provide a filtering for modules that support the `subpath` argument. In case you see any issue with the previous addition, it could be possible to add a disclaimer.

matifali:
[...]

PR #316, opened 2024-10-08
feat(jupyterlab): add support for `subdomain=false`

framctr:
Resolves [#313](https://github.com/coder/modules/issues/313)


framctr:
I changed also the prints in the `run.sh` to be compatible with `/bin/sh` according to https://www.shellcheck.net/wiki/SC2028

jupyterlab/main.tf
```diff
@@ -49,17 +58,26 @@ resource "coder_script" "jupyterlab" {
...
     PORT : var.port
+    BASE_URL : var.subdomain ? "http://localhost:${var.port}/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyterlab" : ""
   })
   run_on_start = true
 }
 
 resource "coder_app" "jupyterlab" {
   agent_id     = var.agent_id
-  slug         = "jupyterlab"
+  slug         = "jupyterlab" # sync with with end of subdomain URL
```
matifali, line 68:
```suggestion
  slug         = "jupyterlab" # sync with the usage in URL
```

jupyterlab/main.tf
```diff
@@ -49,17 +58,26 @@ resource "coder_script" "jupyterlab" {
...
-}
+
+  dynamic "healthcheck" {
+    for_each = var.subdomain ? toset([true]) : toset([])
+    content {
+      url       = "http://localhost:${var.port}/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyterlab"
+      interval  = 6
+      threshold = 10
+    }
+  }
```
matifali, line 83: It looks like it will only work if `subdomain=true` while the `url` is for the case when `subdomain=false`.

Can you please check and also set the healthcheck for both cases?

jupyterlab/run.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env sh
 
+if [ -n "$BASE_URL" ]
```
matifali, line 3:
```suggestion
if [ -n "${BASE_URL}" ]
```
I think, by default, the `$BASE_URL` will be empty for both cases as `terraform` only templates when it is `${BASE_URL}`.

jupyterlab/run.sh
```diff
@@ -15,11 +20,17 @@ if ! command -v jupyterlab > /dev/null 2>&1; then
   fi
   # install jupyterlab
   pipx install -q jupyterlab
-  echo "🥳 jupyterlab has been installed\n\n"
+  printf "%s\n\n" "🥳 jupyterlab has been installed"
```
matifali, line 23: Any motivation behind changing `echo` to `printf`.

matifali (commented):
Thank you for the contribution. :)

jupyterlab/run.sh
matifali: ok. Thanks.

jupyterlab/main.tf
framctr: Unfortunately I cannot since I do not have control on our DNS and I do not have time to set it up a new environment privately..

In case, can we keep as it was before this PR, without the healthcheck?

jupyterlab/run.sh
framctr: Yes you're right

jupyterlab/run.sh
```diff
@@ -15,11 +20,17 @@ if ! command -v jupyterlab > /dev/null 2>&1; then
...
 else
-  echo "🥳 jupyterlab is already installed\n\n"
+  printf "%s\n\n" "🥳 jupyterlab is already installed"
 fi
 
 echo "👷 Starting jupyterlab in background..."
 echo "check logs at ${LOG_PATH}"
```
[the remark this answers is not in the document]
framctr: These are working in any shell and I think it could be convenient anyway to keep as they are

jupyterlab/main.tf
framctr: done

jupyterlab/run.sh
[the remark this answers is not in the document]
framctr: Done

jupyterlab/main.tf
matifali: Yes, let us drop the healthcheck for now.

jupyterlab/main.tf
```diff
@@ -49,17 +58,26 @@ resource "coder_script" "jupyterlab" {
...
```
matifali, line 83:
```suggestion
```

matifali (changes requested):

matifali (approved):

jupyterlab/main.tf
```diff
@@ -49,17 +58,17 @@ resource "coder_script" "jupyterlab" {
...
   run_on_start = true
 }
 
 resource "coder_app" "jupyterlab" {
   agent_id     = var.agent_id
-  slug         = "jupyterlab"
+  slug         = "jupyterlab" # sync with the usage in URL
   display_name = "JupyterLab"
-  url          = "http://localhost:${var.port}"
+  url          = var.subdomain ? "http://localhost:${var.port}/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyterlab" : "http://localhost:${var.port}"
```
matifali, line 70:
```suggestion
  url          = var.subdomain ? "http://localhost:${var.port}" : "http://localhost:${var.port}/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyterlab"
```

jupyterlab/main.tf
```diff
@@ -36,6 +39,12 @@ variable "share" {
   }
 }
 
+variable "subdomain" {
+  type        = bool
+  description = "Determines whether JupyterLab will be accessed via it's own subdomain or whether it will be accessed via a path on Coder."
```
code-asher, line 44: Minor typo:

```suggestion
  description = "Determines whether JupyterLab will be accessed via its own subdomain or whether it will be accessed via a path on Coder."
```

jupyterlab/main.tf
```diff
@@ -49,17 +58,18 @@ resource "coder_script" "jupyterlab" {
   script = templatefile("${path.module}/run.sh", {
     LOG_PATH : var.log_path,
     PORT : var.port
+    BASE_URL : var.subdomain ? "http://localhost:${var.port}/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyterlab" : ""
```
code-asher, line 61: Is this reversed?  With a subdomain, there should be no path, and with a subdomain there should be a path?

matifali, line 61:
```suggestion
    BASE_URL : var.subdomain ? "" : "http://localhost:${var.port}/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyterlab"
```
I think this should be reverse too.

code-asher: Also does this needs to be just the path?

```suggestion
    BASE_URL : var.subdomain ? "" : "/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyterlab""
```

Because:

```
$ jupyter-lab --no-browser --ServerApp.base_url=http://localhost:3000/foo/bar
[I 2024-10-15 18:37:55.987 ServerApp] Jupyter Server 2.14.2 is running at:
[I 2024-10-15 18:37:55.987 ServerApp] http://localhost:8888/http://localhost:8888/foo/bar/lab?token=50a96431d3bfedeef0a1f656fa5382b0bc07a9ebf5086a79

 ```

File: jupyterlab/run.sh [hunk 10]
```diff
@@ -15,11 +19,17 @@ if ! command -v jupyterlab > /dev/null 2>&1; then
...
+  printf "%s\n\n" "🥳 jupyterlab is already installed"
 fi
 
-echo "👷 Starting jupyterlab in background..."
-echo "check logs at ${LOG_PATH}"
-$HOME/.local/bin/jupyter-lab --ServerApp.ip='0.0.0.0' --ServerApp.port=${PORT} --no-browser --ServerApp.token='' --ServerApp.password='' > ${LOG_PATH} 2>&1 &
+printf "👷 Starting jupyterlab in background..."
+printf "check logs at ${LOG_PATH}"
+$HOME/.local/bin/jupyter-lab --no-browser \
+  "$BASE_URL" \
```
code-asher, line 30: Very minor nit, could we rename this to `BASE_URL_FLAG` or something?  I was confused at first because I thought we were putting the base URL directly in the command line invocation.  :laughing:

code-asher: Or we use `--ServerApp.base_url=${BASE_URL}` directly here.  I think a blank string is valid.

jupyterlab/main.tf
```diff
@@ -49,17 +58,18 @@ resource "coder_script" "jupyterlab" {
...
   run_on_start = true
 }
 
 resource "coder_app" "jupyterlab" {
   agent_id     = var.agent_id
-  slug         = "jupyterlab"
+  slug         = "jupyterlab" # sync with the usage in URL
   display_name = "JupyterLab"
-  url          = "http://localhost:${var.port}"
+  url          = var.subdomain ? "http://localhost:${var.port}" : "http://localhost:${var.port}/@${data.coder_workspace_owner.me.name}/${data.coder_workspace.me.name}/apps/jupyterlab"
```
code-asher, line 70: I have not had a chance to run it, this looks good to me at a glance but I am not 100% sure how the proxy works when there is a path in the url.  Has someone ran it and can confirm this works?

jupyterlab/run.sh
matifali: Usually the default value is `/` and not an empty string for some other tools I have seen. Like filebrowser and MS code-server.

jupyterlab/main.tf
matifali: The OP @framctr has tested this in #313

jupyterlab/run.sh
code-asher: `/` makes sense to me!

jupyterlab/main.tf
code-asher: Awesome!  The main reason I asked is that it looked like the flag was incorrect, but it looks like in the issue they used the right value.

matifali:
Hi @framctr, do you plan to finish this? If not, I can help complete this. Thank you :)

framctr:
@matifali
As I said previously, I cannot verify with `subdomain=true` and for this reason, I didn't update anything more. If you can verify this it would be great. 😊

matifali:
I left a few comments and suggestions on possible issues with url.

PR #319, opened 2024-10-14
chore: explicitly setup terraform

matifali:
CI started failing due to runner images missing the terraform.
I decided to use the same terraform version we use in coder/coder.


johnstcn (approved):
stamp 👍

PR #320, opened 2024-10-15
fix(filebrowser): support custom base_url in case of custom db path

### LINKED ISSUE #309: bug(filebrowser): Base URL Configuration Issue in FileBrowser Module
### Problem Description

There's an issue in the FileBrowser module version `1.0.19`, specifically in the [filebrowser/run.sh#L21](https://github.com/coder/modules/blob/438c9045673c629e95416960c94a6b6344e3d298/filebrowser/run.sh#L21) script. The script attempts to customize the server Base URL before starting the server in the background. However, this customization only works when the database is located at the default path, `~/filebrowser.db`.

If the database location has been modified (e.g.: `database_path = "/home/coder/.filebrowser/filebrowser.db"`), the script tries to update the default database location, resulting in no changes. Consequently, the `Base URL` remains empty.

### Current Behavior

```bash
# Current implementation
filebrowser config set --baseurl "${SERVER_BASE_PATH}" > ${LOG_PATH} 2>&1
```

This line only works correctly with the default database location.

### Proposed Solution

To address this issue, we can leverage the existing mechanism used for starting the server. Here's the proposed change:

```diff
# set baseurl to be able to run if sudomain=false; if subdomain=true the SERVER_BASE_PATH value will be ""
-filebrowser config set --baseurl "${SERVER_BASE_PATH}" > ${LOG_PATH} 2>&1
+filebrowser config set --baseurl "${SERVER_BASE_PATH}"$${DB_FLAG} > ${LOG_PATH} 2>&1
```

This modification could ensures that the Base URL is correctly set regardless of the database location (to be tested).

yandolfat:
Proposal to fix #309.

### DIFF
--- a/filebrowser/README.md
+++ b/filebrowser/README.md
@@ -14,7 +14,7 @@ A file browser for your workspace.
 ```tf
 module "filebrowser" {
   source   = "registry.coder.com/modules/filebrowser/coder"
-  version  = "1.0.20"
+  version  = "1.0.22"
   agent_id = coder_agent.example.id
 }
 ```
@@ -28,7 +28,7 @@ module "filebrowser" {
 ```tf
 module "filebrowser" {
   source   = "registry.coder.com/modules/filebrowser/coder"
-  version  = "1.0.20"
+  version  = "1.0.22"
   agent_id = coder_agent.example.id
   folder   = "/home/coder/project"
 }
@@ -39,7 +39,7 @@ module "filebrowser" {
 ```tf
 module "filebrowser" {
   source        = "registry.coder.com/modules/filebrowser/coder"
-  version       = "1.0.20"
+  version       = "1.0.22"
   agent_id      = coder_agent.example.id
   database_path = ".config/filebrowser.db"
 }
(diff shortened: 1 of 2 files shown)


matifali:
Hi @yandolfat, do you plan to come back to this PR? I would be happy to complete this if you are not available. Thanks for the PR.

link-webcreations:
> Hi @yandolfat, do you plan to come back to this PR? I would be happy to complete this if you are not available. Thanks for the PR.

Hi @matifali what would you need me to do? Just rebase from main or something else?

matifali:
@yandolfat Could you test the updated functionality by setting the source as 

```tf
module "filebrowser" {
  source     = "git::https://github.com/yandolfat/modules.git//filebrowser?ref=patches/filebrowser"
  agent_name = "main"
  subdomain  = false
}
```

yandolfat:
> ```hcl
> "git::https://github.com/yandolfat/modules.git//filebrowser?ref=patches/filebrowser"
> ```

Seems to work :)

https://imgur.com/a/YTpEWY5

matifali (approved):

PR #322, opened 2024-10-17
feat(jetbrains-gateway): add slug variable

### LINKED ISSUE #321: jetbrains-gateway: allow specifying slug for multiple use of the module
When trying to add 2 separate buttons - Goland and Pycharm, using <https://registry.coder.com/modules/jetbrains-gateway|jetbrains-gateway> module, user is running into `duplicate app slug, they must be unique per template: "gateway".` 

Could we make the slug a module variable? Or have some other way to allow 2+ instances of `jetbrains-gateway` module per workspace?

[Slack Message](https://codercom.slack.com/archives/C04SK827849/p1729114845226099?thread_ts=1729114845.226099&cid=C04SK827849)

matifali:
- Update with new slug variable to allow using the module multiple times. Resolves #321

### DIFF
--- a/jetbrains-gateway/README.md
+++ b/jetbrains-gateway/README.md
@@ -14,7 +14,7 @@ This module adds a JetBrains Gateway Button to open any workspace with a single 
 ```tf
 module "jetbrains_gateway" {
   source         = "registry.coder.com/modules/jetbrains-gateway/coder"
-  version        = "1.0.20"
+  version        = "1.0.21"
   agent_id       = coder_agent.example.id
   agent_name     = "example"
   folder         = "/home/coder/example"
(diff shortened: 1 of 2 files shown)


dannykopping (approved):
LGTM

_Nit: the reordering should probably have been done in a separate PR (but thank you for doing it!) since it isn't related to this change, and complicated the review._

djarbz:
@matifali I just saw this on your v1.0.21 release.
Just wondering if it might have been better to have the default be the selected IDE? That way it should always be unique by default, I don't see a reason to add the same IDE more than once.

djarbz:
Ahh, I was referring to this default.

```
variable "slug" { 
   type        = string 
   description = "The slug for the coder_app. Allows resuing the module with the same template." 
   default     = "gateway" 
 } 
 
```

Issue #325, opened 2024-10-19
KasmVNC - Doesn't Start - Can't install dependencies


djarbz:
Logs
```
cp: cannot stat '/etc/zsh/newuser.zshrc.recommended': No such file or directory
Detected Distribution: ubuntu
Detected Version: 24.04
Detected Architecture: x86_64
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 2607k  100 2607k    0     0  4513k      0 --:--:-- --:--:-- --:--:-- 4513k
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 kasmvncserver : Depends: libgbm1 (>= 12.0.0~0) but it is not installable
                 Depends: libgl1 but it is not installable
                 Depends: libgomp1 (>= 6) but it is not installable
                 Depends: libxfont2 (>= 1:2.0.1) but it is not installable
                 Depends: libxshmfence1 but it is not installable
                 Depends: xauth but it is not installable
                 Depends: libswitch-perl but it is not installable
E: Unable to correct problems, you have held broken packages.
                 Depends: libyaml-tiny-perl but it is not installable
                 Depends: libhash-merge-simple-perl but it is not installable
                 Depends: liblist-moreutils-perl but it is not installable
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
info: Adding user `coder' to group `ssl-cert' ...
bash: line 1: /etc/kasmvnc/kasmvnc.yaml: No such file or directory
bash: line 173: vncpasswd: command not found
🚀 Starting KasmVNC server...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[...]

PR #326, opened 2024-10-19
fix(kasmVNC): fix debian installation and improve logging

djarbz:
I needed to add an `apt-get update` so that it could install dependencies.

I also took the opportunity to cleanup the log output for both `curl` and `apt-get`.

Additionally, we now log the KASM version.

Fixes https://github.com/coder/modules/issues/325


matifali (approved):
Thank you. @djarbz

djarbz:
Absolutely!

I've noticed an issue with the concurrency of modules with APT locks.

I have a thought about how to possibly address module dependencies, would you prefer I open an issue or discussion?

PR #329, opened 2024-10-24
fix(kasmvnc): optimize KasmVNC deployment script

### LINKED ISSUE #327: kasmVNC: support kasm official images
https://kasmweb.com/docs/latest/guide/custom_images.html

This is what happened when I tried to start my workspace with `kasmweb/postman:1.16.0`:

```sh
[0;1mLogging into Coder...

[0;1mInstalling filebrowser 

⚙️ Creating settings file...
[0;1mInstalling code-server!
Detected Distribution: ubuntu
Detected Version: 20.04
Detected Architecture: x86_64
vncserver is already installed.
vncserver already installed. Skipping installation.
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
🚀 Starting KasmVNC server...
Attempting to authenticate with flag URL: 'https://dev.coder.com'
  Welcome to Coder, bpmct! You're authenticated.
Downloading File Browser for linux/amd64...
https://github.com/filebrowser/filebrowser/releases/download/v2.31.2/linux-amd64-filebrowser.tar.gz
Extracting...
Putting filemanager in /usr/local/bin (may require password)
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
Aborted, error 1 in command: $sudo_cmd mv "$PREFIX/tmp/$filemanager_bin" "$install_path/$filemanager_bin"
🥳 Installation complete! 

👷 Starting filebrowser in background... 

📂 Serving /home/kasm-user at http://localhost:13339 

Running 'filebrowser --noauth --root /home/kasm-user --port 13339' 

📝 Logs at /tmp/filebrowser.log 

#=#=#                                                                         

                                                                           0.0%
#                                                                          2.4%
#############                                                             19.1%
##########################                                                37.3%
#######################################                                   54.8%
###################################################                       72.1%
################################################################          89.4%
######################################################################## 100.0%
🥳 code-server has been installed in /tmp/code-server

jq is required to install extensions from a workspace file.
🥳 code-server has been installed in /tmp/code-server

jq is required to install extensions from a workspace file.
```

djarbz:
I took the opportunity to majorly refactor the run.sh script.

Basically a rewrite from the ground up, and I found a few things that I don't think were tested for non-Ubuntu distros.

I updated the Kasm Version reference to avoid confusion in the script.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/main.tf#L42-L46

Removed the extra shebang and added an early exit signal in case of failure.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/run.sh#L1-L4

Optimize the download function to set the command and then one block of logic to perform the download and handle a failure.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/run.sh#L16-L37

Created a cross platform function to add a user to a group based on which command is available.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/run.sh#L39-L52

Refactored the `install_deb` function to dynamically update the (stale) package cache and wait for a lock.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/run.sh#L54-L69

Check and fail early if sudo or /etc/os-release are not available.

Source `/etc/os-release` since it is formatted as a `.env` file.
Also, adjust the distro and version for Oracle.
Note we are using `distro_version` to make it more readable vs just version.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/run.sh#L100-L108

Cleaned up the ARCH mappings to make it more readable.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/run.sh#L115-L127

For installation, set a base URL that is the same for all distros.
Cleaned up the cases and merged ones that could be merged.
Reduced checking for specific distro versions, we want this to be compatible with future versions without modifying the script each time. If KASM doesn't support the version, then it will fail during the attempt to download.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/run.sh#L129-L154

Finally, added a check to verify that the VNC server was actually running.
https://github.com/coder/modules/blob/a8cc861e831d796914f965ec5c6c5b1b59aa9118/kasmvnc/run.sh#L177-L182

Also closes #327

### DIFF
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -58,8 +58,10 @@ add_user_to_group() {
 # Function to install kasmvncserver for debian-based distros
 install_deb() {
   local url=$1
-  download_file "$url" /tmp/kasmvncserver.deb
-  # Define the directory to check
+  local kasmdeb="/tmp/kasmvncserver.deb"
+
+  download_file "$url" "$kasmdeb"
+  
   CACHE_DIR="/var/lib/apt/lists/partial"
   # Check if the directory exists and was modified in the last 60 minutes
   if [[ ! -d "$CACHE_DIR" ]] || ! find "$CACHE_DIR" -mmin -60 -print -quit &> /dev/null; then
@@ -67,25 +69,57 @@ install_deb() {
     # Update package cache with a 300-second timeout for dpkg lock
     sudo apt-get -o DPkg::Lock::Timeout=300 -qq update
   fi
-  DEBIAN_FRONTEND=noninteractive sudo apt-get -o DPkg::Lock::Timeout=300 install --yes -qq --no-install-recommends --no-install-suggests /tmp/kasmvncserver.deb
+
+  DEBIAN_FRONTEND=noninteractive sudo apt-get -o DPkg::Lock::Timeout=300 install --yes -qq --no-install-recommends --no-install-suggests "$kasmdeb"
+  rm "$kasmdeb"
+
   add_user_to_group "$USER" ssl-cert
-  rm /tmp/kasmvncserver.deb
 }
 
 # Function to install kasmvncserver for rpm-based distros
 install_rpm() {
   local url=$1
-  download_file "$url" /tmp/kasmvncserver.rpm
-  sudo dnf localinstall /tmp/kasmvncserver.rpm
-  rm /tmp/kasmvncserver.rpm
+  local kasmrpm="/tmp/kasmvncserver.rpm"
+  local package_manager
+
+  if command -v dnf &> /dev/null; then
+    # shellcheck disable=SC2034
+    package_manager=(dnf localinstall -y)
+  elif command -v zypper &> /dev/null; then
+    # shellcheck disable=SC2034
+    package_manager=(zypper install -y)
+  elif command -v yum &> /dev/null; then
+    # shellcheck disable=SC2034
+    package_manager=(yum localinstall -y)
+  elif command -v rpm &> /dev/null; then
+    # Do we need to manually handle missing dependencies?
+    # shellcheck disable=SC2034
+    package_manager=(rpm -i)
+  else
+    echo "ERROR: No supported package manager available (dnf, zypper, yum, or rpm required)"
+    exit 1
+  fi
+
+  download_file "$url" "$kasmrpm"
+
+  # shellcheck disable=SC2288
+  sudo "$${package_manager[@]}" "$kasmrpm" || {
+    echo "ERROR: Failed to install $kasmrpm"
+    exit 1
+  }
+
+  rm "$kasmrpm"
 }
 
 # Function to install kasmvncserver for Alpine Linux
 install_alpine() {
   local url=$1
-  download_file "$url" /tmp/kasmvncserver.tgz
-  tar -xzf /tmp/kasmvncserver.tgz -C /usr/local/bin/
-  rm /tmp/kasmvncserver.tgz
+  local kasmtgz="/tmp/kasmvncserver.tgz"
+
+  download_file "$url" "$kasmtgz"
+  
+  tar -xzf "$kasmtgz" -C /usr/local/bin/
+  rm "$kasmtgz"
 }
 
 # Detect system information
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -40,21 +40,6 @@ download_file() {
   }
 }
 
-# Add user to group using available commands
-add_user_to_group() {
-  local user="$1"
-  local group="$2"
-
-  if command -v usermod &> /dev/null; then
-    sudo usermod -aG "$group" "$user"
-  elif command -v adduser &> /dev/null; then
-    sudo adduser "$user" "$group"
-  else
-    echo "ERROR: At least one of 'adduser' or 'usermod' is required"
-    exit 1
-  fi
-}
-
 # Function to install kasmvncserver for debian-based distros
 install_deb() {
   local url=$1
@@ -72,8 +57,6 @@ install_deb() {
 
   DEBIAN_FRONTEND=noninteractive sudo apt-get -o DPkg::Lock::Timeout=300 install --yes -qq --no-install-recommends --no-install-suggests "$kasmdeb"
   rm "$kasmdeb"
-
-  add_user_to_group "$USER" ssl-cert
 }
 
 # Function to install kasmvncserver for rpm-based distros
@@ -117,7 +100,7 @@ install_alpine() {
   local kasmtgz="/tmp/kasmvncserver.tgz"
 
   download_file "$url" "$kasmtgz"
-  
+
   tar -xzf "$kasmtgz" -C /usr/local/bin/
   rm "$kasmtgz"
 }
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -181,10 +181,11 @@ echo -e "password\npassword\n" | vncpasswd -wo -u "$USER"
 # Start the server
 printf "🚀 Starting KasmVNC server...\n"
 vncserver -select-de "${DESKTOP_ENVIRONMENT}" -disableBasicAuth > /tmp/kasmvncserver.log 2>&1 &
+pid=$!
 
 # Wait for server to start
 sleep 5
-if ! pgrep -f vncserver > /dev/null; then
+if ps -p $pid | grep -q "^$pid"; then
   echo "ERROR: Failed to start KasmVNC server. Check logs at /tmp/kasmvncserver.log"
   grep -v '^[[:space:]]*$' /tmp/kasmvncserver.log
   exit 1
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -76,7 +76,7 @@ install_deb() {
 install_rpm() {
   local url=$1
   download_file "$url" /tmp/kasmvncserver.rpm
-  sudo rpm -i /tmp/kasmvncserver.rpm
+  sudo dnf localinstall /tmp/kasmvncserver.rpm
   rm /tmp/kasmvncserver.rpm
 }
 
@@ -116,7 +116,7 @@ echo "Detected Architecture: $arch"
 case "$arch" in
   x86_64)
     if [[ "$distro" =~ ^(ubuntu|debian|kali)$ ]]; then
-    	arch="amd64"
+      arch="amd64"
     fi
     ;;
   aarch64 | arm64)
@@ -161,7 +161,7 @@ else
   echo "vncserver already installed. Skipping installation."
 fi
 
-cat <<EOF > "$HOME/.vnc/kasmvnc.yaml"
+cat << EOF > "$HOME/.vnc/kasmvnc.yaml"
 network:
   protocol: http
   websocket_port: ${PORT}
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -186,7 +186,7 @@ else
   SUDO=
 
   echo "WARNING: Sudo access not available, using user config dir!"
-  
+
   if [[ -f "$kasm_config_file" ]]; then
     echo "WARNING: Custom user KasmVNC config exists, not overwriting!"
     echo "WARNIGN: Ensure that you manually configure the appropriate settings."
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -50,7 +50,7 @@ add_user_to_group() {
   elif command -v adduser &> /dev/null; then
     sudo adduser "$user" "$group"
   else
-    echo "ERROR: At least one of 'adduser'(Debian) 'usermod'(RHEL) is required"
+    echo "ERROR: At least one of 'adduser' or 'usermod' is required"
     exit 1
   fi
 }
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -99,6 +99,8 @@ arch="$(uname -m)"
 if [[ "$ID" == "ol" ]]; then
   distro="oracle"
   distro_version="$${distro_version%%.*}"
+elif [[ "$ID" == "fedora" ]]; then
+  distro_version="$(grep -oP '\(\K[\w ]+' /etc/fedora-release | tr '[:upper:]' '[:lower:]' | tr -d ' ')"
 fi
 
 echo "Detected Distribution: $distro"
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -161,7 +161,7 @@ else
   echo "vncserver already installed. Skipping installation."
 fi
 
-tee "$HOME/.vnc/kasmvnc.yaml" > /dev/null << EOF
+cat <<EOF > "$HOME/.vnc/kasmvnc.yaml"
 network:
   protocol: http
   websocket_port: ${PORT}
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -84,12 +84,6 @@ install_alpine() {
   rm /tmp/kasmvncserver.tgz
 }
 
-# Check for sudo (required)
-if ! command -v sudo &> /dev/null; then
-  echo "ERROR: Required command 'sudo' not found"
-  exit 1
-fi
-
 # Detect system information
 if [[ ! -f /etc/os-release ]]; then
   echo "ERROR: Cannot detect OS: /etc/os-release not found"
@@ -128,6 +122,12 @@ esac
 
 # Check if vncserver is installed, and install if not
 if ! check_installed; then
+  # Check for sudo (required)
+  if ! command -v sudo &> /dev/null; then
+    echo "ERROR: Required command 'sudo' not found"
+    exit 1
+  fi
+
   base_url="https://github.com/kasmtech/KasmVNC/releases/download/v${KASM_VERSION}"
 
   echo "Installing KASM version: ${KASM_VERSION}"
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -46,7 +46,7 @@ install_deb() {
   local kasmdeb="/tmp/kasmvncserver.deb"
 
   download_file "$url" "$kasmdeb"
-  
+
   CACHE_DIR="/var/lib/apt/lists/partial"
   # Check if the directory exists and was modified in the last 60 minutes
   if [[ ! -d "$CACHE_DIR" ]] || ! find "$CACHE_DIR" -mmin -60 -print -quit &> /dev/null; then
@@ -148,7 +148,7 @@ esac
 # Check if vncserver is installed, and install if not
 if ! check_installed; then
   # Check for NOPASSWD sudo (required)
-  if ! command -v sudo &>/dev/null || ! sudo -n true 2>/dev/null; then
+  if ! command -v sudo &> /dev/null || ! sudo -n true 2> /dev/null; then
     echo "ERROR: sudo NOPASSWD access required!"
     exit 1
   fi
@@ -178,7 +178,7 @@ else
   echo "vncserver already installed. Skipping installation."
 fi
 
-if command -v sudo &>/dev/null && sudo -n true 2>/dev/null; then
+if command -v sudo &> /dev/null && sudo -n true 2> /dev/null; then
   kasm_config_file="/etc/kasmvnc/kasmvnc.yaml"
   SUDO=sudo
 else
@@ -190,7 +190,7 @@ else
 fi
 
 echo "Writing KasmVNC config to $kasm_config_file"
-$SUDO tee "$kasm_config_file" > /dev/null <<EOF
+$SUDO tee "$kasm_config_file" > /dev/null << EOF
 network:
   protocol: http
   websocket_port: ${PORT}
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -188,4 +188,4 @@ if ! pgrep -f vncserver > /dev/null; then
   echo "ERROR: Failed to start KasmVNC server. Check logs at /tmp/kasmvncserver.log"
   exit 1
 fi
-printf "🚀 Starting KasmVNC server started successfully!\n"
+printf "🚀 KasmVNC server started successfully!\n"
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -147,9 +147,9 @@ esac
 
 # Check if vncserver is installed, and install if not
 if ! check_installed; then
-  # Check for sudo (required)
-  if ! command -v sudo &> /dev/null; then
-    echo "ERROR: Required command 'sudo' not found"
+  # Check for NOPASSWD sudo (required)
+  if ! command -v sudo &>/dev/null || ! sudo -n true 2>/dev/null; then
+    echo "ERROR: sudo NOPASSWD access required!"
     exit 1
   fi
 
@@ -178,7 +178,19 @@ else
   echo "vncserver already installed. Skipping installation."
 fi
 
-cat << EOF > "$HOME/.vnc/kasmvnc.yaml"
+if command -v sudo &>/dev/null && sudo -n true 2>/dev/null; then
+  kasm_config_file="/etc/kasmvnc/kasmvnc.yaml"
+  SUDO=sudo
+else
+  kasm_config_file="$HOME/.vnc/kasmvnc.yaml"
+  SUDO=
+  mkdir -p "$HOME/.vnc"
+  echo "WARNING: Sudo access not available, using user config dir!"
+  echo "WARNING: This may prevent custom user KasmVNC settings from applying!"
+fi
+
+echo "Writing KasmVNC config to $kasm_config_file"
+$SUDO tee "$kasm_config_file" > /dev/null <<EOF
 network:
   protocol: http
   websocket_port: ${PORT}
--- a/kasmvnc/run.sh
+++ b/kasmvnc/run.sh
@@ -58,7 +58,7 @@ install_deb() {
   # Define the directory to check
   CACHE_DIR="/var/lib/apt/lists/partial"
   # Check if the directory exists and was modified in the last 60 minutes
-  if [ ! -d "$CACHE_DIR" ] || ! find "$CACHE_DIR" -mmin -60 -print -quit &>/dev/null; then
+  if [ ! -d "$CACHE_DIR" ] || ! find "$CACHE_DIR" -mmin -60 -print -quit &> /dev/null; then
     echo "Stale Package Cache, updating..."
     # Update package cache with a 300-second timeout for dpkg lock
     sudo apt-get -o DPkg::Lock::Timeout=300 -qq update
@@ -117,7 +117,7 @@ case "$arch" in
   x86_64)
     [[ "$distro" =~ ^(ubuntu|debian|kali)$ ]] && arch="amd64" || arch="x86_64"
     ;;
-  aarch64|arm64)
+  aarch64 | arm64)
     [[ "$distro" =~ ^(ubuntu|debian|kali)$ ]] && arch="arm64" || arch="aarch64"
     ;;
   *)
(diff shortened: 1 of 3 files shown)


kasmvnc/run.sh
```diff
@@ -14,143 +15,137 @@ check_installed() {
...
+  download_file "$url" /tmp/kasmvncserver.tgz
   tar -xzf /tmp/kasmvncserver.tgz -C /usr/local/bin/
   rm /tmp/kasmvncserver.tgz
 }
 
+# Check for sudo (required)
+if ! command -v sudo &> /dev/null; then
+  echo "ERROR: Required command 'sudo' not found"
+  exit 1
+fi
```
matifali, line 91: I am testing this with https://kasmweb.com/docs/latest/guide/custom_images.html images, which do not have sudo access for the `kasm-user` container user but have kasmvnc pre-installed.

A WIP can be seen in the [`support-kasm-images`](https://github.com/coder/modules/tree/atif/support-kasm-images) branch.

What I did was to check if `vncserver` command is already available and then skip the installation and directly move to starting the vncserver.

matifali (approved):
The refactor LGTM. And yes we did not tested enough for all non ubuntu distros.

kasmvnc/run.sh
```diff
@@ -159,21 +154,29 @@ else
 fi
 
 # Coder port-forwarding from dashboard only supports HTTP
```
matifali, line 156:
```suggestion
```
> [!IMPORTANT]
>Unrelated to your PR, but while we are the refactor,

This line is redundant, as Coder now supports http and https port forwarding. 
Also, the `protocol : http` is one of the `vnc|http` options and doe snot have to do anything with https.

kasmvnc/run.sh
```diff
@@ -159,21 +154,29 @@ else
...
+sudo tee /etc/kasmvnc/kasmvnc.yaml > /dev/null << EOF
 network:
   protocol: http
   websocket_port: ${PORT}
   ssl:
     require_ssl: false
   udp:
     public_ip: 127.0.0.1
-EOF"
+EOF
```
matifali, line 165: For images without `sudo`
we can also write the same file in `$HOME/.vnc/kasmvnc.yaml`, and it overrides the one in `/etc/kasmvnc/kasmvnc.yaml` if one exists.

kasmvnc/run.sh
```diff
@@ -159,21 +154,29 @@ else
...
-sudo -u $USER bash -c "vncserver -select-de ${DESKTOP_ENVIRONMENT} -disableBasicAuth" > /tmp/kasmvncserver.log 2>&1 &
+# shellcheck disable=SC2024
+sudo -u "$USER" bash -c "vncserver -select-de ${DESKTOP_ENVIRONMENT} -disableBasicAuth" > /tmp/kasmvncserver.log 2>&1 &
+
+# Wait for server to start
+sleep 5
+if ! pgrep -f vncserver > /dev/null; then
+  echo "ERROR: Failed to start KasmVNC server. Check logs at /tmp/kasmvncserver.log"
+  exit 1
+fi
```
matifali, line 182: This is super helpful. 👍🏼

kasmvnc/run.sh
```diff
@@ -159,21 +154,29 @@ else
...
 # The server is protected via the Coder session token / tunnel
 # and does not listen publicly
-echo -e "password\npassword\n" | vncpasswd -wo -u $USER
+echo -e "password\npassword\n" | vncpasswd -wo -u "$USER"
 
 # Start the server
 printf "🚀 Starting KasmVNC server...\n"
-sudo -u $USER bash -c "vncserver -select-de ${DESKTOP_ENVIRONMENT} -disableBasicAuth" > /tmp/kasmvncserver.log 2>&1 &
+# shellcheck disable=SC2024
+sudo -u "$USER" bash -c "vncserver -select-de ${DESKTOP_ENVIRONMENT} -disableBasicAuth" > /tmp/kasmvncserver.log 2>&1 &
```
matifali, line 175: Handle for the case when there is no `sudo`. All `coder_script` resources are run by the user who runs the `coder_agent.*.init_script`, and it may not have passwordless sudo or no sudo at all. 

We can make these changes in a separate PR, too, but I recommend we handle them now, as this is a major refactor.

djarbz: Do we need sudo to start the `vncserver`?

kasmvnc/run.sh
djarbz: Should we just default to using the home directory?

kasmvnc/run.sh
djarbz: Can we remove these lines then?
```yaml
ssl:
  require_ssl: false
```

kasmvnc/run.sh
matifali: As far as I have tested no.

kasmvnc/run.sh
matifali: Probably yes. But for the official kasm images, I have observed they bake a kasm.yaml that requires SSL certs, so I had to extend/patch the image.

kasmvnc/run.sh
matifali: If we remove this, the kasmvncserver requires specifying SSL certs. We do not need those as the connection is already secured by Coder.

kasmvnc/run.sh
djarbz: I saw that issue after posting this PR. I tried to run postman with these changes, but the workspace never finished starting up.
But, if we write our version to the home directory, then it should always use that one and shouldn't matter what they baked in?

djarbz: What if we change it to:
```yaml
network:
  protocol: http
  websocket_port: ${PORT}
  ssl:
    require_ssl: false
    pem_certificate:
    pem_key:
  udp:
    public_ip: 127.0.0.1
```

djarbz:
@matifali I have pushed a few changes and just tested locally.
We now only need sudo if we are installing kasm.
We are running KasmVNC as the coder user.

Still need to test with non-Ubuntu images, but I don't have any readily available.
Will also need to test if this works for @bpmct and his issue with official KasmWeb images.

kasmvnc/run.sh
matifali: This works 🎉 . I thought blank would be an invalid value. But it looks like it's still a valid string.

matifali:
@djarbz I just tested with `kasmweb/postman:1.16.0,`, and VNC starts, although I do not see a postman icon on the Desktop.

djarbz:
> @djarbz I just tested with `kasmweb/postman:1.16.0,`, and VNC starts, although I do not see a postman icon on the Desktop.

Is it available in the application menu?
The $HOME directory might be overridden or somehow masked, or maybe it never existed on the desktop?

djarbz:
@matifali Can you share your template?
I just tried with a minimal template and I get stuck on the dashboard loading.

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
+echo -e "password\npassword\n" | vncpasswd -wo -u "$USER"
 
 # Start the server
 printf "🚀 Starting KasmVNC server...\n"
-sudo -u $USER bash -c "vncserver -select-de ${DESKTOP_ENVIRONMENT} -disableBasicAuth" > /tmp/kasmvncserver.log 2>&1 &
+vncserver -select-de "${DESKTOP_ENVIRONMENT}" -disableBasicAuth > /tmp/kasmvncserver.log 2>&1 &
+
+# Wait for server to start
+sleep 5
+if ! pgrep -f vncserver > /dev/null; then
```
mafredri, line 179: `pgrep` isn't the most portable, isn't installed by default on e.g. Debian images. Another option is to store the pid and check it:

```
vncserver ... &
pid=$!

sleep 5
if ps -p $pid | grep -q "^$pid"; then
```

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
+  echo "ERROR: Failed to start KasmVNC server. Check logs at /tmp/kasmvncserver.log"
+  exit 1
+fi
+printf "🚀 Starting KasmVNC server started successfully!\n"
```
mafredri, line 183:
```suggestion
printf "🚀 KasmVNC server started successfully!\n"
```

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
 
 # Start the server
 printf "🚀 Starting KasmVNC server...\n"
-sudo -u $USER bash -c "vncserver -select-de ${DESKTOP_ENVIRONMENT} -disableBasicAuth" > /tmp/kasmvncserver.log 2>&1 &
+vncserver -select-de "${DESKTOP_ENVIRONMENT}" -disableBasicAuth > /tmp/kasmvncserver.log 2>&1 &
+
+# Wait for server to start
+sleep 5
+if ! pgrep -f vncserver > /dev/null; then
+  echo "ERROR: Failed to start KasmVNC server. Check logs at /tmp/kasmvncserver.log"
```
mafredri, line 180: Should we print last few lines from log for ease of debugging?

mafredri (commented):
Thanks for putting so much effort into this refactor, much appreciated. Left some suggestions and questions.

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
-      fi
-      ;;
-    centos)
-      if [[ "$version" == "7" ]]; then
-        install_rpm_centos7 "https://github.com/kasmtech/KasmVNC/releases/download/v${VERSION}/kasmvncserver_centos_core_${VERSION}_$${arch}.rpm"
-      else
-        install_rpm "https://github.com/kasmtech/KasmVNC/releases/download/v${VERSION}/kasmvncserver_centos_core_${VERSION}_$${arch}.rpm"
-      fi
+    oracle | fedora | opensuse)
+      bin_name="kasmvncserver_$${distro}_$${distro_version}_${KASM_VERSION}_$${arch}.rpm"
```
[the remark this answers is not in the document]
djarbz: Good call, working on a commit address this.

```bash
$ grep -oP '\(\K[\w ]+' /etc/fedora-release | tr '[:upper:]' '[:lower:]' | tr -d ' '
thirtynine
```

kasmvnc/run.sh
```diff
@@ -14,166 +15,167 @@ check_installed() {
...
+  download_file "$url" /tmp/kasmvncserver.tgz
   tar -xzf /tmp/kasmvncserver.tgz -C /usr/local/bin/
   rm /tmp/kasmvncserver.tgz
 }
 
 # Detect system information
-distro=$(grep "^ID=" /etc/os-release | awk -F= '{print $2}')
-version=$(grep "^VERSION_ID=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
-arch=$(uname -m)
+if [[ ! -f /etc/os-release ]]; then
```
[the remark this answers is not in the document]
djarbz: I would believe so, but I cannot confirm 100%.
That being said, the original script was referencing this file, so all I changed was how we were interacting with that information.

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
 
-# Function to install kasmvncserver for debian-based distros
-install_deb() {
-  local url=$1
-  download_file $url /tmp/kasmvncserver.deb
-  sudo apt-get update
-  DEBIAN_FRONTEND=noninteractive sudo apt-get install --yes -qq --no-install-recommends --no-install-suggests /tmp/kasmvncserver.deb
-  sudo adduser $USER ssl-cert
-  rm /tmp/kasmvncserver.deb
+  $download_tool "$url" > "$output" || {
```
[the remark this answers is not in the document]
djarbz: I'll give that bash array idea a try here shortly.

We do only call it once and only if we need to install the server, if the server is already installed, this is skipped.

kasmvnc/run.sh
djarbz: I just checked, the log file is fairly short, we can probably print the whole thing.

```
$ cat /tmp/kasmvncserver.log 

New 'qlcplus-v4-workspace:1 (coder)' desktop is qlcplus-v4-workspace:1

Users configured:
coder (can use keyboard and mouse, add/remove users)

Log file is /home/coder/.vnc/qlcplus-v4-workspace:1.log

Starting applications specified in /home/coder/.vnc/xstartup

Paste this url in your browser:
https://10.88.0.67:6800
```

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
+  download_file "$url" /tmp/kasmvncserver.deb
+  # Define the directory to check
+  CACHE_DIR="/var/lib/apt/lists/partial"
+  # Check if the directory exists and was modified in the last 60 minutes
+  if [ ! -d "$CACHE_DIR" ] || ! find "$CACHE_DIR" -mmin -60 -print -quit &> /dev/null; then
+    echo "Stale Package Cache, updating..."
+    # Update package cache with a 300-second timeout for dpkg lock
+    sudo apt-get -o DPkg::Lock::Timeout=300 -qq update
+  fi
+  DEBIAN_FRONTEND=noninteractive sudo apt-get -o DPkg::Lock::Timeout=300 install --yes -qq --no-install-recommends --no-install-suggests /tmp/kasmvncserver.deb
```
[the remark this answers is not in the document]
djarbz: I'm not sure, I left that in from the original script.
It's working for me, so as of right now, I don't think anything important is missing.

kasmvnc/run.sh
```diff
@@ -14,166 +15,167 @@ check_installed() {
...
+  elif command -v adduser &> /dev/null; then
+    sudo adduser "$user" "$group"
+  else
+    echo "ERROR: At least one of 'adduser'(Debian) 'usermod'(RHEL) is required"
+    exit 1
+  fi
 }
 
-# Function to install kasmvncserver for CentOS 7
-install_rpm_centos7() {
```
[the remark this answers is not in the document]
djarbz: I wasn't looking at the documentation for the install, but rather the release page.
There is no release for Centos, at least not for v1.3.2.
I do see it was compiled for v1.3.1.

I'm assuming that their builds are fully automated, and they are dropping support for Centos?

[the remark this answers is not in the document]
mafredri: It doesn't look like they've intentionally dropped CentOS, but can't say what's going on there. They never added support for CentOS 8 even though an issue has been open since 2020. And "just" last year they fixed a build issue with CentOS 7, and no updates or removals since then `¯\_(ツ)_/¯`.

It's also still in the official docs: https://www.kasmweb.com/kasmvnc/docs/latest/install.html#centos-oracle-7

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
+  add_user_to_group "$USER" ssl-cert
+  rm /tmp/kasmvncserver.deb
 }
 
 # Function to install kasmvncserver for rpm-based distros
 install_rpm() {
   local url=$1
-  download_file $url /tmp/kasmvncserver.rpm
+  download_file "$url" /tmp/kasmvncserver.rpm
   sudo rpm -i /tmp/kasmvncserver.rpm
```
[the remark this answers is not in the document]
djarbz: We should look into finding or building a desktop image to test this with for RHEL, Fedora, Centos, and Oracle.
As with the original script I believe only the `codercom/enterprise-desktop:ubuntu` image was tested and this is the one I am using.

I did some further reading, it sounds like on RHEL8, [yum is an alias for dnf](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/8.4_release_notes/distribution-of-content-in-rhel-8#package_management_with_yum_dnf).

> On Red Hat Enterprise Linux 8, installing software is ensured by the YUM tool, which is based on the DNF technology. We deliberately adhere to usage of the yum term for consistency with previous major versions of RHEL. However, if you type dnf instead of yum, the command works as expected because yum is an alias to dnf for compatibility.

Other places suggest that dnf is preferred almost always to yum for the extra features.

I've also read that we should use `sudo dnf localinstall /tmp/kasmvncserver.rpm` instead.

kasmvnc/run.sh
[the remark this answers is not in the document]
djarbz: I think we can ignore it for now and add support if someone needs it?
I would imagine most people are going to user the enterprise-desktop image anyways.

Either that or an official KasmWeb image, which is still a WIP.

kasmvnc/run.sh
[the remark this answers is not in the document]
djarbz: I think we can leave it as is and wait for an issue if someone runs into an image where this is not available.
However, I'm not sure where we would gather the required information then.

kasmvnc/run.sh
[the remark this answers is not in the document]
djarbz: Works great in my test.

kasmvnc/run.sh
djarbz: Added and tested.

kasmvnc/run.sh
djarbz: I added this to the exit block, wondering if we want to just print this by default after the sleep?
There is some good information in here and it is short.

mafredri: I worry it could be _a lot_ longer if there were a slew of errors, but it's hard to predict. I'd probably do:
```
echo "Showing the last 5 log lines, inspect the full log at /tmp/kasmvncserver.log:"
tail -n5 /tmp/kasmvncserver.log
```
But I wouldn't reject the PR if you decide it's better to print it all. 😄

mafredri: > I added this to the exit block, wondering if we want to just print this by default after the sleep? There is some good information in here and it is short.

That's an interesting idea, feel free to play around with it.

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
+add_user_to_group() {
+  local user="$1"
+  local group="$2"
+
+  if command -v usermod &> /dev/null; then
+    sudo usermod -aG "$group" "$user"
+  elif command -v adduser &> /dev/null; then
+    sudo adduser "$user" "$group"
+  else
+    echo "ERROR: At least one of 'adduser'(Debian) 'usermod'(RHEL) is required"
```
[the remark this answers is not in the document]
djarbz: I have removed the distro verbiage.

kasmvnc/run.sh
djarbz: It looks good on a successful run, not sure what an error will show, but I limited it to 10 non-blank lines.

```
🚀 Starting KasmVNC server...
New 'qlcplus-v4-workspace:1 (coder)' desktop is qlcplus-v4-workspace:1
Users configured:
coder (can use keyboard and mouse, add/remove users)
Log file is /home/coder/.vnc/qlcplus-v4-workspace:1.log
Starting applications specified in /home/coder/.vnc/xstartup
Paste this url in your browser:
https://10.88.0.72:6800
🚀 KasmVNC server started successfully!
```

kasmvnc/run.sh
```diff
@@ -14,166 +15,179 @@ check_installed() {
...
-      ;;
-    centos)
-      if [[ "$version" == "7" ]]; then
-        install_rpm_centos7 "https://github.com/kasmtech/KasmVNC/releases/download/v${VERSION}/kasmvncserver_centos_core_${VERSION}_$${arch}.rpm"
-      else
-        install_rpm "https://github.com/kasmtech/KasmVNC/releases/download/v${VERSION}/kasmvncserver_centos_core_${VERSION}_$${arch}.rpm"
-      fi
+    oracle | fedora | opensuse)
+      bin_name="kasmvncserver_$${distro}_$${distro_version}_${KASM_VERSION}_$${arch}.rpm"
+      install_rpm "$base_url/$bin_name"
```
mafredri, line 149: We probably need both `install_rpm` and `install_dnf`. I just tried the `opensuse/leap` image and it has neither `dnf` nor `yum`, only `zypper` and `rpm`.

I'm not sure if this is a script meant for installing the kasmvnc built package or something else, but this at least gives hints for the expectation per distro: https://github.com/kasmtech/KasmVNC/blob/3a8517d7dc461eaccc7ed8e3d3b155e233426fc8/builder/scripts/common.sh#L22-L29

kasmvnc/run.sh
```diff
@@ -14,166 +15,169 @@ check_installed() {
...
-else
-  echo "Unsupported architecture: $arch"
-  exit 1
-fi
+case "$arch" in
+  x86_64)
+    [[ "$distro" =~ ^(ubuntu|debian|kali)$ ]] && arch="amd64" || arch="x86_64"
+    ;;
+  aarch64 | arm64)
+    [[ "$distro" =~ ^(ubuntu|debian|kali)$ ]] && arch="arm64" || arch="aarch64"
```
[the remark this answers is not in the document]
mafredri: Doesn't look like `arm64` case in `aarch64 | arm64)` is ever applicable:

```
❯ for container in ubuntu debian alpine fedora opensuse/leap container-registry.oracle.com/os/oraclelinux:9-slim; do for plat in linux/arm64 linux/amd64; do docker run -it --rm --platform $plat $container /bin/sh -c 'echo $(grep ^NAME= /etc/os-release): $(uname -m)'; done; done 2>/dev/null
NAME="Ubuntu": aarch64
NAME="Ubuntu": x86_64
NAME="Debian GNU/Linux": aarch64
NAME="Debian GNU/Linux": x86_64
NAME="Alpine Linux": aarch64
NAME="Alpine Linux": x86_64
NAME="Fedora Linux": aarch64
NAME="Fedora Linux": x86_64
NAME="openSUSE Leap": aarch64
NAME="openSUSE Leap": x86_64
NAME="Oracle Linux Server": aarch64
NAME="Oracle Linux Server": x86_64
```

So we can essentially use the same logic as above for `x86_64`, default=aarch64, for debian based, use arm64 instead (verified from kasm releases that all others are aarch64).

kasmvnc/run.sh
```diff
@@ -14,166 +15,179 @@ check_installed() {
...
       exit 1
       ;;
   esac
 else
   echo "vncserver already installed. Skipping installation."
 fi
 
-# Coder port-forwarding from dashboard only supports HTTP
-sudo bash -c "cat > /etc/kasmvnc/kasmvnc.yaml <<EOF
+cat << EOF > "$HOME/.vnc/kasmvnc.yaml"
```
mafredri, line 164: I just realized we're unconditionally overwriting the users kasmvnc configuration here.

We should switch back to writing this to `/etc` since it will still apply to the user. That'll allow the user to provide their own settings in dotfiles.

> KasmVNC is configured via YAML based configurations. The server level configuration is at /etc/kasmvnc/kasmvnc.yaml. **Edits to this file apply to all users**. Individual users can override server global configurations by specifying them in their configuration file at ~/.vnc/kasmvnc.yaml.

(emphasis mine.)

mafredri (commented):
Thanks for making all the amends and thinking of alternative solutions, this is starting to look good, just a few more issues remain.

kasmvnc/run.sh
djarbz: Perhaps in the `install_rpm` function, we have a switch case to select based on what tooling is available?
Similar to the download function?

Priority:
1. dnf
2. zypper
3. yum
4. rpm

kasmvnc/run.sh
djarbz: Regarding arch mapping.

Is this what you are thinking?
```bash
case "$arch" in
  x86_64)
    if [[ "$distro" =~ ^(ubuntu|debian|kali)$ ]]; then
      arch="amd64"
    fi
    ;;
  aarch64)
    if [[ "$distro" =~ ^(ubuntu|debian|kali)$ ]]; then
      arch="arm64"
    fi
    ;;
  arm64)
    :  # This is effectively a noop
    ;;
  *)
    echo "ERROR: Unsupported architecture: $arch"
    exit 1
    ;;
esac
```

kasmvnc/run.sh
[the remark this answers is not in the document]
djarbz: FYI, doesn't look like anything important.
```
Suggested packages:
  pciutils
Recommended packages:
  libgl1-amber-dri mesa-vulkan-drivers | vulkan-icd
```

kasmvnc/run.sh
mafredri: > Regarding arch mapping.

Yes, except in my testing the noop for `arm64` isn't necessary at all.

> My concern with the /etc kasmvnc.yaml is the requirement for sudo and when using the official KasmWeb images, they have some settings preconfigured.

In what situations can KasmVNC be installed if sudo isn't available? I suppose KasmWeb images are the only case since it's preinstalled?

I think explicit overwrite of config in /etc is fine, they're using our module for a reason. Down the line we may want to think about introducing more KasmVNC settings and that means we again want to overwrite the yaml again. Merging yaml seems like a cumbersome process requiring additional tools for manipulation.

If we fallback to writing the config in user home, we should check it's presence and only write if it's missing. If we detect an existing config we can print a warning that if it doesn't work, consider deleting the file.

djarbz: >Yes, except in my testing the noop for arm64 isn't necessary at all.
I'm just thinking of future, don't want to fail out if it is reported as such?

If we use a KasmWeb image, we do not have sudo access.
This means that we cannot write to the /etc config file and must write to the home directory.
I think we should default to /etc if sudo is available, but fallback to /home with a warning message in case the user is trying to figure out why their /home config might not be taking effect.

djarbz: I have added a commit to handle this with /etc as default and a warning if we need to fallback to /home for the config.

kasmvnc/run.sh
```diff
@@ -14,166 +15,216 @@ check_installed() {
...
       ;;
   esac
 else
   echo "vncserver already installed. Skipping installation."
 fi
 
-# Coder port-forwarding from dashboard only supports HTTP
-sudo bash -c "cat > /etc/kasmvnc/kasmvnc.yaml <<EOF
+if command -v sudo &> /dev/null && sudo -n true 2> /dev/null; then
+  kasm_config_file="/etc/kasmvnc/kasmvnc.yaml"
```
mafredri, line 182: Do we need `sudo mkdir -p /etc/kasmvnc` here like for the user config?

kasmvnc/run.sh
```diff
@@ -14,166 +15,216 @@ check_installed() {
...
+else
+  kasm_config_file="$HOME/.vnc/kasmvnc.yaml"
+  SUDO=
+
+  echo "WARNING: Sudo access not available, using user config dir!"
+
+  if [[ -f "$kasm_config_file" ]]; then
+    echo "WARNING: Custom user KasmVNC config exists, not overwriting!"
+    echo "WARNING: Ensure that you manually configure the appropriate settings."
+    kasm_config_file="/dev/stderr"
```
mafredri, line 193: Nice solution still outputting the file 👍🏻

I think this is great for now. This could be enhanced in the future by validating the config (e.g. grep "websocket_port: ...") and if it's detected as invalid, it could be backed up (rename) and then a new valid config written in its place. Let's not do any of that in this PR though, it's already big enough. 😄

mafredri (approved):
Awesome work, thanks! I had one last question but short of validating the script on all platforms, I have nothing more to add, approved!

mafredri:
Looks like package-lock was accidentally included in this PR, could you remove it?

kasmvnc/run.sh
djarbz: I have not run into the need for it.
I can add it as a catchall just in case if you want?

kasmvnc/run.sh
djarbz: I agree, there is a lot we "can" do, but what is truly necessary, especially at this point.

PR #334, opened 2024-10-28
feat(filebrowser): check if already installed

IamTaoChen:
Add check if the `filebrowser` exist

### DIFF
--- a/filebrowser/run.sh
+++ b/filebrowser/run.sh
@@ -1,11 +1,21 @@
 #!/usr/bin/env bash
 
 BOLD='\033[0;1m'
-printf "$${BOLD}Installing filebrowser \n\n"
 
-curl -fsSL https://raw.githubusercontent.com/filebrowser/get/master/get.sh | bash
+# Check if filebrowser is installed
+if command -v filebrowser &> /dev/null; then
+    printf "🥳 Filebrowser is already installed. Skipping installation.\n\n$"
+else
+    printf "${BOLD}Installing Filebrowser...\n\n"
 
-printf "🥳 Installation complete! \n\n"
+    # Install Filebrowser
+    if curl -fsSL https://raw.githubusercontent.com/filebrowser/get/master/get.sh | bash; then
+        printf "🥳 Installation complete! Filebrowser is now installed.\n\n"
+    else
+        printf "❌ Installation failed! Please check the logs.\n\n"
+        exit 1
+    fi
+fi
 
 printf "👷 Starting filebrowser in background... \n\n"
 
--- a/filebrowser/run.sh
+++ b/filebrowser/run.sh
@@ -6,7 +6,7 @@ BOLD='\033[0;1m'
 if command -v filebrowser &> /dev/null; then
     printf "🥳 Filebrowser is already installed. Skipping installation.\n\n$"
 else
-    printf "${BOLD}Installing Filebrowser...\n\n"
+    printf "$${BOLD}Installing Filebrowser...\n\n"
 
     # Install Filebrowser
     if curl -fsSL https://raw.githubusercontent.com/filebrowser/get/master/get.sh | bash; then
--- a/filebrowser/run.sh
+++ b/filebrowser/run.sh
@@ -2,20 +2,14 @@
 
 BOLD='\033[0;1m'
 
-# Check if filebrowser is installed
-if command -v filebrowser &> /dev/null; then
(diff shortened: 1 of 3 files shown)


filebrowser/run.sh
```diff
@@ -1,11 +1,21 @@
...
-printf "$${BOLD}Installing filebrowser \n\n"
 
-curl -fsSL https://raw.githubusercontent.com/filebrowser/get/master/get.sh | bash
-
-printf "🥳 Installation complete! \n\n"
+# Check if filebrowser is installed
+if command -v filebrowser &> /dev/null; then
+    printf "🥳 Filebrowser is already installed. Skipping installation.\n\n$"
+else
+    printf "${BOLD}Installing Filebrowser...\n\n"
```
matifali, line 9:
```suggestion
    printf "$${BOLD}Installing Filebrowser...\n\n"
```
${} are templates by terraform so we need $$ for shell variables.

IamTaoChen:
Done

matifali:
Hi @IamTaoChen could you also run `bun install && bun fmt`?
Then run `bun test filebrowser/main.test.ts` with docker installed to run tests for the module. Make sure they pass.

IamTaoChen:
During testing, I encountered an error in main.tf related to the variable validation logic:
```bash
modules2 git:(main) ✗ bun test filebrowser/main.test.ts --verbose
bun test v1.1.33 (247456b6)

filebrowser/main.test.ts:
Terraform encountered problems during initialisation, including problems
with the configuration, described below.

The Terraform configuration must be valid before initialization so that
Terraform can determine which modules and providers need to be installed.
╷
│ Error: Invalid reference in variable validation
│ 
│   on main.tf line 27, in variable "agent_name":
│   27:     condition     = var.subdomain || var.agent_name != ""
│ 
│ The condition for variable "agent_name" can only refer to the variable itself, using var.agent_name.
╵
```
Terraform’s validation rules don’t allow a variable’s validation logic to reference other variables, such as `var.subdomain`. If I comment out or mask this part of the code, the test passes successfully. However, I’m unsure if masking it in this PR is the right approach.

Should I proceed by temporarily masking the validation to unblock the tests, or would you recommend another way to handle this dependency.

matifali:
@IamTaoChen Let us remove the `validation` from `agent_name` and make it an optional variable with a default set to `""`.

IamTaoChen:
Now, it passed the test

filebrowser/run.sh
```diff
@@ -1,9 +1,13 @@
 #!/usr/bin/env bash
 
 BOLD='\033[0;1m'
+
 printf "$${BOLD}Installing filebrowser \n\n"
 
-curl -fsSL https://raw.githubusercontent.com/filebrowser/get/master/get.sh | bash
+# Check if filebrowser is installed
+if ! command -v filebrowser &> /dev/null; then
```
mafredri, line 8: (No action, just commentary.)

This is nice, but unfortunately it's rare that users will benefit since it's usually installed to `/usr/local/bin`. For containers it's rarely persisted.

It'd be cool if we could store this in `$CODER_SCRIPT_BIN_DIR` instead. Usually this is within `/tmp` but it's possible to change by starting the coder agent with `CODER_AGENT_SCRIPT_DATA_DIR=/path/to/persistent/dir`.

This is something to consider for all modules, should we track this in a new issue @matifali?

mafredri (approved):
LGTM, thanks for your contribution!

filebrowser/run.sh
matifali: > This is nice, but unfortunately it's rare that users will benefit since it's usually installed to /usr/local/bin. For containers it's rarely persisted.

For air-gapped scenarios where the image already has it installed this is beneficial. 

> It'd be cool if we could store this in $CODER_SCRIPT_BIN_DIR instead. Usually this is within /tmp but it's possible to change by starting the coder agent with CODER_AGENT_SCRIPT_DATA_DIR=/path/to/persistent/dir.

I am a bit confused about $CODER_SCRIPT_BIN_DIR and $CODER_AGENT_SCRIPT_DATA_DIR. Do we have them already and how are the used? Or are you proposing this new concept.

IamTaoChen:
Hi I just noticed that `/icon/filebrowser.svg` doesn't exist on version of 2.16

filebrowser/run.sh
matifali: This is super useful and we should start using it in our modules.

PR #341, opened 2024-11-18
chore: add health check badge

matifali:


.github/workflows/check.yaml
```diff
@@ -1,5 +1,5 @@
-name: Check modules on registry.coder.com
-
+name: Health
```
matifali, line 1: the name change seems to be required to get a shorter name for the health check badge.

johnstcn (approved):

PR #342, opened 2024-11-18
chore: integrate Instatus in check script

matifali:
Add Instatus API support to monitor module status.

### DIFF
--- a/.github/scripts/check.sh
+++ b/.github/scripts/check.sh
@@ -2,6 +2,21 @@
 set -o pipefail
 set -u
 
+# List of required environment variables
+required_vars=(
+    "INSTATUS_API_KEY"
+    "INSTATUS_PAGE_ID"
+    "INSTATUS_COMPONENT_ID"
+)
+
+# Check if each required variable is set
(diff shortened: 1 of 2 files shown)


.github/scripts/check.sh
```diff
@@ -1,24 +1,70 @@
...
+
+# Function to update the component status on Instatus
+update_component_status() {
+    local component_status=$1
+    # see https://instatus.com/help/api/components
+    (curl -X PUT "https://api.instatus.com/v1/$INSTATUS_PAGE_ID/components/$INSTATUS_COMPONENT_ID" \
+        -H "Authorization: Bearer $INSTATUS_API_KEY" \
+        -H "Content-Type: application/json" \
+        -d "{\"status\": \"$component_status\"}")
+}
```
johnstcn, line 26: Can we check for the required InStatus environment variables being set upfront and exit with a non-zero exit status if not provided? Right now we will exit with

```
./.github/scripts/check.sh: line 22: INSTATUS_PAGE_ID: unbound variable

```

johnstcn (approved):

PR #346, opened 2024-11-22
Improve incident management in Instatus check script

matifali:
- Change incident naming for better clarity.
- Ensure new incidents are not created if unresolved ones exist.


Parkreiner (commented):
The Bash seemed straightforward. The only thing I'm wondering about (and this might be because I don't know InStatus yet, and haven't dealt with the old version of the registry much) is whether there would be situations when we would want multiple incidents? 

As in, if a service is experiencing multiple problems, wouldn't we want reports on each category of problem we're experiencing? If I'm understanding the current code right, if one problem gets reported, that also silences any future problems, even if they're not directly related

I'm definitely out of my element here, though. If you think that's not a problem, I can go ahead and approve

matifali:
> As in, if a service is experiencing multiple problems, wouldn't we want reports on each category of problem we're experiencing? If I'm understanding the current code right, if one problem gets reported, that also silences any future problems, even if they're not directly related


Yes, you are right. Currently, the flow is to create a new incident when the registry is down every 15 minutes. I am here trying to prevent that and update an existing one. But yes there is a chance that subsequent events could be unrelated.

InStatus has a very flexible API that we can explore to improve this flow and should probably move into the rgsitry repo.

PR #348, opened 2024-11-22
chore: add updates to force redeployment on Vercel

Parkreiner:
## Changes made
- Updated `check.sh` script to add support for automatic re-deploying in the event that the the registry has a partial/full outage.

### DIFF
--- a/.github/scripts/check.sh
+++ b/.github/scripts/check.sh
@@ -18,8 +18,8 @@ for var in "${required_vars[@]}"; do
     fi
 done
 
-LATEST_REDEPLOY_FAILED="$LATEST_REDEPLOY_FAILED:-0"
-if (LATEST_REDEPLOY_FAILED); do
+LATEST_REDEPLOY_FAILED="${LATEST_REDEPLOY_FAILED:-0}"
+if [[ "${LATEST_REDEPLOY_FAILED}" -ne "0" ]] ;then
     echo "Trying to re-run job when previous re-deploy failed"
     return 1
 fi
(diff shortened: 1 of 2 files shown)


.github/scripts/check.sh
```diff
@@ -17,6 +19,11 @@ for var in "${required_vars[@]}"; do
     fi
 done
 
+JUST_REDEPLOYED="$JUST_REDEPLOYED:-0"
+if (JUST_REDEPLOYED); do
+    return 1
```
bcpeinhardt, line 24: Maybe an echo for context here

.github/scripts/check.sh
```diff
@@ -108,6 +152,20 @@ else
...
+    # ASAP
+    status_code=$(force_redeploy_registry)
+    # shellcheck disable=SC2181
+    if (( status_code == 200 )); then
+        echo "Reployment successful"
+    else
+        echo "Unable to redeploy automatically"
+    fi
+
+    # Update environment variable so that if automatic re-deployment fails, we
```
bcpeinhardt, line 166: Update this comment to specify a successful redeploy != things are all fixed.

johnstcn (approved):
made some changes and shellcheck no longer complains, so it "should" be fine 🤞

johnstcn (approved):

.github/scripts/check.sh
```diff
@@ -17,14 +23,20 @@ for var in "${required_vars[@]}"; do
     fi
 done
 
+LATEST_REDEPLOY_FAILED="${LATEST_REDEPLOY_FAILED:-0}"
```
bcpeinhardt, line 26: We can remove this

bcpeinhardt (approved):

PR #350, opened 2024-11-23
ci: revert schedule changes

johnstcn:


johnstcn:
it did the thing on its own finally! https://github.com/coder/modules/actions/runs/11982294380

PR #355, opened 2024-11-27
feat(jetbrains-gateway): bump to 2024.3

matifali:
1. Bumps default IDE versions to 2024.3
2. Set the default value of `latest=true` to always fetch the latest versions.

Extracted from #339 as that introduces a breaking change.

### DIFF
--- a/jetbrains-gateway/README.md
+++ b/jetbrains-gateway/README.md
@@ -14,7 +14,7 @@ This module adds a JetBrains Gateway Button to open any workspace with a single 
 ```tf
 module "jetbrains_gateway" {
   source         = "registry.coder.com/modules/jetbrains-gateway/coder"
-  version        = "1.0.23"
+  version        = "1.0.25"
   agent_id       = coder_agent.example.id
   agent_name     = "example"
   folder         = "/home/coder/example"
@@ -32,7 +32,7 @@ module "jetbrains_gateway" {
 ```tf
 module "jetbrains_gateway" {
   source         = "registry.coder.com/modules/jetbrains-gateway/coder"
-  version        = "1.0.23"
+  version        = "1.0.25"
   agent_id       = coder_agent.example.id
   agent_name     = "example"
   folder         = "/home/coder/example"
@@ -46,7 +46,7 @@ module "jetbrains_gateway" {
 ```tf
 module "jetbrains_gateway" {
   source         = "registry.coder.com/modules/jetbrains-gateway/coder"
-  version        = "1.0.23"
+  version        = "1.0.25"
   agent_id       = coder_agent.example.id
   agent_name     = "example"
   folder         = "/home/coder/example"
@@ -61,7 +61,7 @@ module "jetbrains_gateway" {
 ```tf
 module "jetbrains_gateway" {
   source         = "registry.coder.com/modules/jetbrains-gateway/coder"
-  version        = "1.0.23"
+  version        = "1.0.25"
   agent_id       = coder_agent.example.id
   agent_name     = "example"
   folder         = "/home/coder/example"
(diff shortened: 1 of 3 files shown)


jetbrains-gateway/README.md
```diff
@@ -41,18 +41,18 @@ module "jetbrains_gateway" {
 }
 ```
 
-### Use the latest release version
+### Use the fixed set by `jetbrains_ide_versions`
```
dannykopping, line 44: Customers won't know what this means.
Either we should say the below (since we're _interpreting_ the stability of a release by setting it as our default), or we should link to the definition.

```suggestion
### Use the latest stable release
```

jetbrains-gateway/main.tf
```diff
@@ -80,36 +80,36 @@ variable "jetbrains_ide_versions" {
   description = "The set of versions for each jetbrains IDE"
   default = {
     "IU" = {
-      build_number = "241.14494.240"
```
dannykopping, line 83: Do we document these changes in any changelog?

dannykopping (approved):

jetbrains-gateway/main.tf
matifali: For now, we write them manually as part of release notes.

https://github.com/coder/modules/releases

jetbrains-gateway/README.md
matifali: As the default behavior is now changed so I wanted to change the example to teach someone on how to use the fixed version passed by `jetbrains_ide_versions`.

matifali: What is your suggestion on how to improve it? We can add a detailed example on using custom versions using the `jetbrains_ide_versions` varibale.

dannykopping: I gave two options here: https://github.com/coder/modules/pull/355#discussion_r1860562310

I think we should allow the user to specify their own `build_number` and `version`, TBH.
If I'm an operator and I know that a certain version is stable, I won't want coder messing that up by always using latest.

phorcys420 (approved):
LGTM except what Danny already pointed out

jetbrains-gateway/README.md
matifali: Yes, then I should just set `stable = false` which stop Coder from messing it up.

matifali: > I think we should allow the user to specify their own build_number and version, TBH.

This is tricky as diffrent JetBrians IDEs share the same `version` but different in `build_number`s.

jetbrains-gateway/README.md
```diff
@@ -41,18 +41,52 @@ module "jetbrains_gateway" {
...
-  version        = "1.0.23"
+  version        = "1.0.25"
   agent_id       = coder_agent.example.id
   agent_name     = "example"
   folder         = "/home/coder/example"
   jetbrains_ides = ["GO", "WS"]
   default        = "GO"
-  latest         = true
+  latest         = false
+  jetbrains_ide_versions = {
```
dannykopping, line 56: Do they have to define all of them?
As an operator I'd probably just care about the one(s) my team uses.

matifali: No they don't have to. They can shorten this to the ones they use.

dannykopping: Let's keep the suggestions realistic then; let's maybe specify IntelliJ+PyCharm and add a comment saying that one or more can be defined?

matifali: Will do.

matifali: updated.

matifali:
@mafredri I need some help with `./update-versions.sh` as it is trying to bump the nested versions too.

jetbrains-gateway/README.md
```diff
@@ -14,7 +14,7 @@ This module adds a JetBrains Gateway Button to open any workspace with a single
 ```tf
 module "jetbrains_gateway" {
   source         = "registry.coder.com/modules/jetbrains-gateway/coder"
-  version        = "1.0.23"
+  version        = "1.0.25"
```
bpmct, line 17: Since we are changing the default behavior, do we want to do a more aggressive version bump?

matifali: Yes. We can do that but that will make all future versions of other modules go to 1.1.X from now on. That would cause confusion for the users of other modules. 

I can also remove the change in default behaviour in this PR and do once we have #157

mafredri (approved):

PR #356, opened 2024-11-28
fix(update-versions.sh): handle markdown/tf block nesting when updating version

mafredri:
The script was too greedy and updating unrelated fields in #355.

This change makes the script much more strict, only updating version on the top-level _iff_ within a markdown code-block.


matifali (approved):
Thank you :ship:

@bcpeinhardt could you give it a review too?

PR #362, opened 2024-12-10
coder_monitoring module

defelmnq:
This PR aims to add a new coder module - named `coder_monitoring` which can be used to enable and customize monitoring of resources over the coder workspaces.

### DIFF
--- a/monitoring/README.md
+++ b/monitoring/README.md
@@ -1,0 +1,72 @@
+---
+display_name: Monitoring
+description: Monitoring of workspace resources
+maintainer_github: coder
+verified: true
+tags: [monitoring]
+---
+
+# Monitoring
+
+This module adds monitoring of workspace resources.
+
+```tf
+module "monitoring" {
+  source  = "registry.coder.com/modules/monitoring/coder"
+  version = "1.0.0"
+  agent_id = coder_agent.dev.id
+}
+```
+
+## Examples
+
+```tf
+module "monitoring" {
+  source  = "registry.coder.com/modules/monitoring/coder"
+  version = "1.0.0"
+  agent_id = coder_agent.dev.id
+}
+```
+
+### Enable/Disable
+
+You can customize the monitoring by setting the `enabled`, `memory_enabled`, and `disk_enabled` variables.
+
+```tf
+module "monitoring" {
+  source  = "registry.coder.com/modules/monitoring/coder"
+  version = "1.0.0"
+  agent_id = coder_agent.dev.id
+  enabled = false
+  memory_enabled = true
+  disk_enabled = false
+}
+```
+
+### Customize Thresholds
+
+You can customize the thresholds by setting the `threshold`, `memory_threshold`, and `disk_threshold` variables.
+
+```tf
+module "monitoring" {
+  source  = "registry.coder.com/modules/monitoring/coder"
+  version = "1.0.0"
+  agent_id = coder_agent.dev.id
+  threshold = 90
+  memory_threshold = 95
+  disk_threshold = 90
+}
+```
+
+### Customize Disks
+
+You can customize the disks by setting the `disks` variable.
+
+```tf
+module "monitoring" {
+  source  = "registry.coder.com/modules/monitoring/coder"
+  version = "1.0.0"
+  agent_id = coder_agent.dev.id
+  disks = ["/"]
+}
+```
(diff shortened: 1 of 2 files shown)


monitoring/main.tf
```diff
@@ -0,0 +1,75 @@
+terraform {
+  required_version = ">= 1.0"
+
+  required_providers {
+    coder = {
+      source  = "coder/coder"
+      version = ">= 0.11"
```
dannykopping, line 7: Why `0.11`? We're on v2.0.2 now.

You'd need to add the `coder_monitoring` datasource to the terraform provider before we can merge this, do you have a PR in the works for this?

monitoring/README.md
```diff
@@ -0,0 +1,72 @@
...
+---
+
+# Monitoring
+
+This module adds monitoring of workspace resources.
+
+```tf
+module "monitoring" {
+  source  = "registry.coder.com/modules/monitoring/coder"
+  version = "1.0.0"
```
dannykopping, line 16: The versioning of modules right now follows the whole repo I believe, so it'd be `v1.0.25`.

monitoring/main.tf
```diff
@@ -0,0 +1,75 @@
...
+variable "threshold" {
+  type        = number
+  description = "The threshold for the monitoring, used for all resources unless overridden by *_threshold."
+  default     = 90
+}
+
+variable "memory_threshold" {
+  type        = number
+  description = "The threshold for the memory monitoring."
+  default     = 90
```
matifali, line 26: Should we specify what units we expect in the description?

monitoring/main.tf
```diff
@@ -0,0 +1,75 @@
...
+}
+
+variable "disk_threshold" {
+  type        = number
+  description = "The threshold for the disk monitoring."
```
matifali, line 31: Could add units.

monitoring/main.tf
```diff
@@ -0,0 +1,75 @@
...
+  default     = 90
+}
+
+variable "disks" {
+  type        = list(string)
+  description = "The disks to monitor."
```
matifali, line 37: Could add an example on how to provide the list.

monitoring/main.tf
```diff
@@ -0,0 +1,75 @@
...
+variable "enabled" {
+  type        = bool
+  description = "Whether the monitoring is enabled."
+  default     = true
+}
+
+variable "memory_enabled" {
+  type        = bool
+  description = "Whether the memory monitoring is enabled."
+  default     = false
```
matifali, line 50: Why disabled? Any reason?

monitoring/main.tf
```diff
@@ -0,0 +1,75 @@
...
+  default     = 90
+}
+
+variable "disks" {
+  type        = list(string)
+  description = "The disks to monitor."
+  default     = ["/"]
+}
+
+variable "enabled" {
```
matifali, line 41: Why do we have a variable for this? 
Can use the terraform `count` to control if the module is added or not.

monitoring/main.tf
```diff
@@ -0,0 +1,75 @@
...
+}
+
+data "coder_monitoring" "monitoring" {
+  name         = "monitoring"
+  description  = var.description
+  threshold    = var.threshold
+  memory_threshold = var.memory_threshold
+  disk_threshold = var.disk_threshold
+  disks = var.disks
+  enabled = var.enabled
```
matifali, line 71: Same observation here. Why do we use enabled argument? We can use the terraform meta argument `count`

matifali (changes requested):

monitoring/main.tf
matifali: Also is there an RFC design on how `coder_monitoring` will work.?

defelmnq:
Hey , thanks for the fast comments.

@matifali I resolved most of your points - just asking a question on the enabled one. ✅

monitoring/main.tf
defelmnq: I imagined we can use this enabled parameter to let users keep the block but just change the enabled variable if they want, for some specific cases, disable it quickly without loosing their configuration, does it make sense ? Otherwise I can delete it and indeed rely on the terraform count (I need to read about it.)

matifali: That's the same purpose the `count` argument can serve. 
It can be set to 0 or 1 to enable or disable the whole module. 
If someone is not using the module they can use the count on the `data "coder_monitoring"` too. 
see: https://developer.hashicorp.com/terraform/language/meta-arguments/count

defelmnq:
Closing until approvals on the design ✅ Still some points to validate together first.

PR #363, opened 2024-12-11
feat: ci to build new registry on push to main

bcpeinhardt:
This PR adds a github actions workflow for deploying new revisions of the registry on pushes to main.
This means updating the new registry will continue to be as simple as landing a PR in this repo, but it should only take as long as the docker container takes to build to see the updates live :) 

I don't *love* that cloud build wants me to pass the build secret and API key as query params but I went through the raw logs for the workflow and it doesn't get logged there.


.github/workflows/deploy-registry.yaml
```diff
@@ -0,0 +1,25 @@
+name: deploy-registry
+
+on:
+  push:
+    branches:
+      - main
+
+  pull_request:
```
matifali, line 8: Do we want to deploy on each PR too?

bcpeinhardt: No I should pull that out, was moving too quickly.

.github/scripts/deploy-registry.sh
```diff
@@ -0,0 +1,37 @@
...
+        exit 1
+    fi
+done
+
+# Trigger a build for dev 
+# DONT EVER SET ANY FLAGS THAT MIGHT PRINT THE URL, AS IT HAS SECRETS IN IT
+curl -X POST "https://cloudbuild.googleapis.com/v1/projects/coder-registry-1/triggers/http-build-registry-v2-dev:webhook?key=${GCLOUD_API_KEY}&secret=${GCLOUD_DEV_DEPLOY_SECRET}" \
+-H "Content-Type: application/json" \
+-d '{}' \
+--fail
```
johnstcn, line 30: I'm not entirely convinced about this approach. We could maybe instead trigger a build using the `gcloud` CLI (ref: https://cloud.google.com/sdk/gcloud/reference/builds/triggers/run) in combination with the `setup-gcloud` action (https://github.com/google-github-actions/setup-gcloud)

.github/scripts/deploy-registry.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Trigger a build for prod
+# DONT EVER SET ANY FLAGS THAT MIGHT PRINT THE URL, AS IT HAS SECRETS IN IT
+curl -X POST "https://cloudbuild.googleapis.com/v1/projects/coder-registry-1/triggers/http-build-registry-v2-trigger:webhook?key=${GCLOUD_API_KEY}&secret=${GCLOUD_PROD_DEPLOY_SECRET}" \
+-H "Content-Type: application/json" \
+-d '{}' \
+--fail
```
johnstcn, line 37: Should this perhaps be a separate step in the workflow?

.github/scripts/deploy-registry.sh
johnstcn: cc @deansheather for some additional eyes.

deansheather: Absolutely would prefer to use gcloud CLI rather than calling googleapis from curl. We also should not be using service account credentials as actions secrets, and should use workload identity instead. I can help you set that up with a new SA for this repo @bcpeinhardt

bcpeinhardt: Secrets deleted, will update to use the gcloud cli, and yes some help with a service account would be great haha 😎

.github/scripts/deploy-registry.sh
bcpeinhardt: Honestly I would even consider *removing* it from the workflow. Since we have a dev environment now a little delay to review there might not be the worst thing.

.github/workflows/deploy-registry.yaml
```diff
@@ -0,0 +1,38 @@
+name: deploy-registry
+
+on:
+  push:
+    branches:
+      - main
+      - bcpeinhardt/ci-to-build-new-registry-on-push-to-main
```
matifali, line 7: don't forget to take this out.
```suggestion
```

matifali (approved):

johnstcn (approved):

bcpeinhardt:
Uh merging in main cause the git-clone test to fail. I don't think that was me but would appreciate a second look.

PR #376, opened 2024-12-20
ci: disable vercel redeploys

johnstcn:


Parkreiner (approved):
Change is small and obvious, but when would we be getting rid of the whole function, rather than commenting it out?

johnstcn:
> Change is small and obvious, but when would we be getting rid of the whole function, rather than commenting it out?

I was thinking next year; doing a vercel redeploy right now would have no effect.

matifali (approved):

PR #377, opened 2024-12-26
chore(git-commit-signing): add version warning in README

phorcys420:
Git versions prior to 2.34 do not support commit signing via an SSH key.

I was willing to add a check in the script, but Git does not return the version as-is (sometimes also returns the commit hash) and we would need to compare versions, I think this would add unnecessary dependencies, what do you think @matifali?
```
coder@catfood> git -v
git version 2.47.0
```


matifali (approved):
This is enough to add to docs. Thanks for taking care of these details.

PR #383, opened 2025-01-08
chore(jetbrains-gateway): make `agent_name` unused

phorcys420:
`agent_name` is not necessary in the URL when the agent ID is provided, this PR aims to remove the need to specify the agent's name to improve the UX of the module.


phorcys420:
Working properly on dogfood with https://dev.coder.com/templates/coder/coder/versions/stupefied_archimedes6

image

jetbrains-gateway/main.test.ts
```diff
@@ -10,19 +10,17 @@ describe("jetbrains-gateway", async () => {
...
   it("should create a link with the default values", async () => {
     const state = await runTerraformApply(import.meta.dir, {
       // These are all required.
       agent_id: "foo",
-      agent_name: "foo",
       folder: "/home/coder",
     });
     expect(state.outputs.url.value).toBe(
-      "jetbrains-gateway://connect#type=coder&workspace=default&owner=default&agent=foo&folder=/home/coder&url=https://mydeployment.coder.com&token=$SESSION_TOKEN&ide_product_code=IU&ide_build_number=243.21565.193&ide_download_link=https://download.jetbrains.com/idea/ideaIU-2024.3.tar.gz",
+      "jetbrains-gateway://connect#type=coder&workspace=default&owner=defaultfolder=/home/coder&url=https://mydeployment.coder.com&token=$SESSION_TOKEN&ide_product_code=IU&ide_build_number=243.21565.193&ide_download_link=https://download.jetbrains.com/idea/ideaIU-2024.3.tar.gz",
```
matifali, line 23:
```suggestion
      "jetbrains-gateway://connect#type=coder&workspace=default&owner=default&folder=/home/coder&url=https://mydeployment.coder.com&token=$SESSION_TOKEN&ide_product_code=IU&ide_build_number=243.21565.193&ide_download_link=https://download.jetbrains.com/idea/ideaIU-2024.3.tar.gz",
```

jetbrains-gateway/main.tf
```diff
@@ -26,7 +26,9 @@ variable "slug" {
 
 variable "agent_name" {
   type        = string
-  description = "Agent name."
+  description = "Agent name. (unused)"
```
matifali, line 29:
```suggestion
  description = "Agent name. (unused). Will be removed in a future version"
```

jetbrains-gateway/main.tf
```diff
@@ -295,8 +297,8 @@ resource "coder_app" "gateway" {
     data.coder_workspace.me.name,
     "&owner=",
     data.coder_workspace_owner.me.name,
-    "&agent=",
-    var.agent_name,
+    #    "&agent=",
+    #    var.agent_name,
```
matifali, line 301:
```suggestion
```
I think it is ok to remove these lines. Even if someone specifies the input it will not be used.

matifali (changes requested):

jetbrains-gateway/main.tf
phorcys420: I did remove it, I just had uncommited changes, sorry!

phorcys420:
@matifali I was wondering, is there a way to add a warning to a template so that the template admins will notice?
Like a `coder_warning` resource or something.

matifali (approved):

matifali:
You can probably hack something using a `null_resource` and `local_exec` to log something if the input is being used. 

I was also thinking about a similar feature to read logs from arbitrary sources and inject them in startup logs. If we do something like that we can expand the scope to choose the sink of the logs to be either build logs (i.e. terraform) or agent logs (i.e.  `startup_script` logs)

code-asher:
Ahh wait we tried this once already, did something change about the ID?  It did not work before because the ID here (of the resource) is not necessarily the same as the ID in the API response (of the agent in the database).

matifali:
> Ahh wait we tried this once already, did something change about the ID? It did not work before because the ID here (of the resource, I think it was changing every rebuild) is not necessarily the same as the ID in the API response (of the agent in the database, this one I think was not changing).

@phorcys420 have you tested by rebuilding/restarting the workspace too?

phorcys420:
> > Ahh wait we tried this once already, did something change about the ID? It did not work before because the ID here (of the resource, I think it was changing every rebuild) is not necessarily the same as the ID in the API response (of the agent in the database, this one I think was not changing).
> 
> @phorcys420 have you tested by rebuilding/restarting the workspace too?

I haven't, I only tried starting it once, i'll check if it works.

PR #384, opened 2025-01-10
chore: set git clone & dotfiles as optional

ericpaulsen:
customers are making good use of our dotfiles & git clone modules. however, there are scenarios where users may not want to set a repo URL or dotfiles URL, despite the module being present in the template.

the changes in this PR prevent the modules from failing if such fields are unset.

### DIFF
--- a/dotfiles/run.sh
+++ b/dotfiles/run.sh
@@ -18,6 +18,6 @@ if [ -n "$${DOTFILES_URI// }" ]; then
 
     CODER_BIN=$(which coder)
     DOTFILES_USER_HOME=$(eval echo ~"$DOTFILES_USER")
-    sudo -u "$DOTFILES_USER" sh -c "'$CODER_BIN' dotfiles '$DOTFILES_URI' -y 2>&1 | tee '$DOTFILES_USER_HOME'/.dotfiles.log"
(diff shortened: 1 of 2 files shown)


phorcys420:
Hello @ericpaulsen, my opinion is that the `count` attribute of the module should be set based on a `coder_parameter` value for this kind of use-case, what do you think?

phorcys420:
I was convinced that something like this would work:
```hcl
data "coder_parameter" "enable_code_server" {
  name    = "code-server?"
  type    = "bool"
  default = true
}

module "code-server" {
  count                   = data.coder_parameter.enable_code_server.value == true ? 1 : 0
  source                  = "dev.registry.coder.com/modules/code-server/coder"
  version                 = ">= 1.0.0"
  agent_id                = coder_agent.dev.id
  folder                  = local.repo_dir
}
```

But I couldn't get it to ever add the module, maybe it doesn't work with modules.
@matifali do you have an idea?

phorcys420:
@ericpaulsen that said, I don't think we should make the script more permissive, rather add a do-nothing toggle if using `count` doesn't work.

matifali:
We should handle this in the terraform and make the input optional and skip running the `coder_script`.

phorcys420:
@matifali were you able to take a look? otherwise i can take it from now on

matifali:
@phorcys420 no I wasn't. It would be nice if you can. Thanks 🙏

matifali:
More context in #469

PR #387, opened 2025-01-24
fix(dotfiles): handle failures in dotfiles installation

chazmo03:
If the dotfiles installation (`coder dotfiles ...`) fails, the script still shows as successful in the Coder UI. This should bubble up any errors so that Coder (and the user) are aware that the installation failed.

### DIFF
--- a/dotfiles/run.sh
+++ b/dotfiles/run.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+set -euo pipefail
+
 DOTFILES_URI="${DOTFILES_URI}"
 DOTFILES_USER="${DOTFILES_USER}"
 
@@ -19,9 +22,5 @@ if [ -n "$${DOTFILES_URI// }" ]; then
     CODER_BIN=$(which coder)
     DOTFILES_USER_HOME=$(eval echo ~"$DOTFILES_USER")
     sudo -u "$DOTFILES_USER" sh -c "'$CODER_BIN' dotfiles '$DOTFILES_URI' -y 2>&1 | tee '$DOTFILES_USER_HOME'/.dotfiles.log"
-  fi
-  if [ $? -ne 0 ]; then
-    echo "Failed to install dotfiles"
-    exit 1
   fi
 fi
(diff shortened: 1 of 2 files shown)


matifali (approved):

matifali:
HI @chazmo03 Thanks for the contribution. Could you please address the failed CI checks?

dotfiles/run.sh
```diff
@@ -20,4 +20,8 @@ if [ -n "$${DOTFILES_URI// }" ]; then
     DOTFILES_USER_HOME=$(eval echo ~"$DOTFILES_USER")
     sudo -u "$DOTFILES_USER" sh -c "'$CODER_BIN' dotfiles '$DOTFILES_URI' -y 2>&1 | tee '$DOTFILES_USER_HOME'/.dotfiles.log"
   fi
+  if [ $? -ne 0 ]; then
+    echo "Failed to install dotfiles"
+    exit 1
+  fi
```
phorcys420, line 26: Thanks for catching this!

I would much rather if we used bash's built-in exception catching mechanism instead.

Could you add `set -euo pipefail` at the beginning of file instead?

fyi:
`-e`: Exit immediately if a simple command exits with a non-zero status
`-u`: Treat unset variables as an error when performing parameter expansion
`-o pipefail`: Set the option corresponding to 'pipefail'. TL;DR exit if anything in a pipe fails

see more at https://ss64.com/bash/set.html

phorcys420 (changes requested):

dotfiles/run.sh
chazmo03: Thanks for the suggestion and explanation @phorcys420. I just pushed that change.

Note that I used the `if [ $? ne 0 ];` style as that's what I found in use in several of the other `run.sh` files.

chazmo03:
@phorcys420 I made the suggested change. Can you please take another look?

phorcys420 (approved):
LGTM!
Sorry for the delay on such a simple PR, I've been pretty busy and away for FOSDEM!

PR #397, opened 2025-02-04
docs: add jetbrains min requirements note and link

EdwardAngert:
resolves https://github.com/coder/coder/discussions/16321

### DIFF
--- a/jetbrains-gateway/README.md
+++ b/jetbrains-gateway/README.md
@@ -11,6 +11,9 @@ tags: [ide, jetbrains, helper, parameter]
 
 This module adds a JetBrains Gateway Button to open any workspace with a single click.
 
+JetBrains recommends a minimum of 4 CPU cores and 8GB of RAM.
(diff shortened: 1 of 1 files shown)


jetbrains-gateway/README.md
```diff
@@ -11,6 +11,9 @@ tags: [ide, jetbrains, helper, parameter]
 
 This module adds a JetBrains Gateway Button to open any workspace with a single click.
 
+JetBrains recommends a minimum of 4 CPU cores and 8GB of RAM.
+Consult the [JetBrains documentation](https://www.jetbrains.com/help/idea/prerequisites.html#min_requirements) to confirm other system requirements.
```
matifali, line 15: I wonder if we should make this highlighted? A warning block maybe?

EdwardAngert: we can - I just try to use them sparingly since it's so easy to ignore highlight blocks sometimes. (and we tend to overuse them, which leads to more ignoring)

PR #398, opened 2025-02-06
fix: handle extensions.json comments in vscode-web and code-server

### LINKED ISSUE #385: vscode-web: auto installing extensions fails if json has comments
When using the `vscode-web` module with `auto_install_extensions` set to true extensions will fail if your repo's `.vscode/extensions.json` file contains comments. 

This appears to be because `run.sh` executes `jq` (which doesn't directly support json with comments) against the file ([see here](https://github.com/coder/modules/blob/main/vscode-web/run.sh#L95)). Comments in `.vscode/extensions.json` are fairly common (for example, see the extensions.json in the [vscode repo](https://github.com/microsoft/vscode/blob/main/.vscode/extensions.json#L2)).

matifali:
Resolves #385

### DIFF
--- a/code-server/run.sh
+++ b/code-server/run.sh
@@ -104,7 +104,8 @@ if [ "${AUTO_INSTALL_EXTENSIONS}" = true ]; then
 
   if [ -f "$WORKSPACE_DIR/.vscode/extensions.json" ]; then
     printf "🧩 Installing extensions from %s/.vscode/extensions.json...\n" "$WORKSPACE_DIR"
-    extensions=$(jq -r '.recommendations[]' "$WORKSPACE_DIR"/.vscode/extensions.json)
+    # Use sed to remove single-line comments before parsing with jq
+    extensions=$(sed 's|//.*||g' "$WORKSPACE_DIR"/.vscode/extensions.json | jq -r '.recommendations[]')
     for extension in $extensions; do
       if extension_installed "$extension"; then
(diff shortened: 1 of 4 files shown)


code-server/run.sh
```diff
@@ -104,7 +104,8 @@ if [ "${AUTO_INSTALL_EXTENSIONS}" = true ]; then
 
   if [ -f "$WORKSPACE_DIR/.vscode/extensions.json" ]; then
     printf "🧩 Installing extensions from %s/.vscode/extensions.json...\n" "$WORKSPACE_DIR"
-    extensions=$(jq -r '.recommendations[]' "$WORKSPACE_DIR"/.vscode/extensions.json)
+    # Use sed to remove single-line comments before parsing with jq
+    extensions=$(sed 's|//.*||g' "$WORKSPACE_DIR"/.vscode/extensions.json | jq -r '.recommendations[]')
```
ThomasK33, line 108: This is sufficient as a quick fix. However, since both scripts will have Node available, we may want to consider parsing and stringifying the string using Node, as that should also eliminate multiline comments and comments after a field or value. (or even use a third-party NPM package to do so).

ThomasK33 (approved):

code-server/run.sh
matifali: That would be great. We can also get rid of `jq` then.

PR #400, opened 2025-02-08
fix(filebrowser): configure before running

### LINKED ISSUE #399: Recent update to the filebrowser moduel breaks subdomains
In the recent commit 22b2ad5feceed89557e5374d1197227e270fb97e the command was changed to fix the fact that the filebrowser.db file didn't exist on first run so the `filebrowser config set --baseurl` command fails until the second time you run it.

However the way the command is interpreted if you use a subdomain and so therefore the baseurl is "" causes the command to fail because it thinks that there is no argument provided to the --baseurl option

```
$ cat /tmp/filebrowser.log
Error: flag needs an argument: --baseurl
Usage:
  filebrowser [flags]
  filebrowser [command]
...
```

Because of this it now fails to start at all when using a subdomain

evilhamsterman:
Fixes #399 

Instead of trying to cram the config into one command we check if the db exists, if not initialize it then configure the various values before starting the app.

### DIFF
--- a/filebrowser/run.sh
+++ b/filebrowser/run.sh
@@ -24,7 +24,7 @@ fi
 # Check if filebrowser db exists
 if [ ! -f ${DB_PATH} ]; then
   filebrowser $DB_FLAG config init >> ${LOG_PATH} 2>&1
-  filebrowser $DB_FLAG config set --baseurl ${SERVER_BASE_PATH} --port ${PORT} --auth.method=noauth --root $ROOT_DIR >> ${LOG_PATH} 2>&1
+  filebrowser $DB_FLAG config set --baseurl=${SERVER_BASE_PATH} --port=${PORT} --auth.method=noauth --root=$ROOT_DIR >> ${LOG_PATH} 2>&1
   filebrowser $DB_FLAG users add admin "" --perm.admin=true >> ${LOG_PATH} 2>&1
 fi
 
--- a/filebrowser/run.sh
+++ b/filebrowser/run.sh
@@ -11,7 +11,7 @@ fi
 
 printf "🥳 Installation complete! \n\n"
 
-printf "👷 Starting filebrowser in background... \n\n"
+printf "🛠️  Configuring filebrowser \n\n"
 
 ROOT_DIR=${FOLDER}
 ROOT_DIR=$${ROOT_DIR/\~/$HOME}
@@ -21,10 +21,20 @@ if [ "${DB_PATH}" != "filebrowser.db" ]; then
   DB_FLAG=" -d ${DB_PATH}"
 fi
 
+# Check if filebrowser db exists
+if [ ! -f ${DB_PATH} ]; then
+  filebrowser $DB_FLAG config init > ${LOG_PATH} 2>&1
+  filebrowser $DB_FLAG config set  --baseurl ${SERVER_BASE_PATH} --port ${PORT} --auth.method=noauth --root $ROOT_DIR > ${LOG_PATH} 2>&1
+  filebrowser $DB_FLAG users add admin "" --perm.admin=true > ${LOG_PATH} 2>&1
+fi
+
+printf "👷 Starting filebrowser in background... \n\n"
+
+
 printf "📂 Serving $${ROOT_DIR} at http://localhost:${PORT} \n\n"
 
-printf "Running 'filebrowser --noauth --root $ROOT_DIR --port ${PORT}$${DB_FLAG} --baseurl ${SERVER_BASE_PATH}' \n\n"
+printf "Running 'filebrowser $DB_FLAG' \n\n"
 
-filebrowser --noauth --root $ROOT_DIR --port ${PORT}$${DB_FLAG} --baseurl ${SERVER_BASE_PATH} > ${LOG_PATH} 2>&1 &
+filebrowser $DB_FLAG > ${LOG_PATH} 2>&1 &
 
 printf "📝 Logs at ${LOG_PATH} \n\n"
--- a/filebrowser/run.sh
+++ b/filebrowser/run.sh
@@ -23,9 +23,9 @@ fi
 
 # Check if filebrowser db exists
 if [ ! -f ${DB_PATH} ]; then
-  filebrowser $DB_FLAG config init > ${LOG_PATH} 2>&1
-  filebrowser $DB_FLAG config set  --baseurl ${SERVER_BASE_PATH} --port ${PORT} --auth.method=noauth --root $ROOT_DIR > ${LOG_PATH} 2>&1
-  filebrowser $DB_FLAG users add admin "" --perm.admin=true > ${LOG_PATH} 2>&1
+  filebrowser $DB_FLAG config init >> ${LOG_PATH} 2>&1
+  filebrowser $DB_FLAG config set --baseurl ${SERVER_BASE_PATH} --port ${PORT} --auth.method=noauth --root $ROOT_DIR >> ${LOG_PATH} 2>&1
+  filebrowser $DB_FLAG users add admin "" --perm.admin=true >> ${LOG_PATH} 2>&1
 fi
 
 printf "👷 Starting filebrowser in background... \n\n"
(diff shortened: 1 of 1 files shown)


matifali:
Hi @evilhamsterman, Could you take care of the tests that are failing?
Also, @dstoffel, can you review and verify if it works for your use case?

Thanks :)

dstoffel:
it's working correctly, thank you

evilhamsterman:
So looking at the failing tests they fail even with the older versions of the script. I have tested all those options myself and they all work. What is the expected scripting environment, the script was using `#!/usr/bin/env bash` for the hash bang and calling bash to execute the script from curl. But the tests appear to be running in the Alpine container which doesn't have bash.

Also it seems something's wrong with how it checks Stdout. Even if I change the container to something like debian:bookworm which has bash and make the script do nothing but `echo "Hello world"` it doesn't register any output in the test and says it just received an empty string

```typescript
28 |     const state = await runTerraformApply(import.meta.dir, {
29 |       agent_id: "foo",
30 |     });
31 |     const output = await executeScriptInContainer(state, "debian:bookworm-slim");
32 |     expect(output.exitCode).toBe(0);
33 |     expect(output.stdout).toEqual([
                               ^
error: expect(received).toEqual(expected)

  [
+   ""
-   "👷 Starting filebrowser in background... ",
-   "",
-   "📂 Serving /root at http://localhost:13339 ",
-   "",
-   "Running 'filebrowser -d filebrowser.db' ",
-   "",
-   "📝 Logs at /tmp/filebrowser.log"
  ]

- Expected  - 7
+ Received  + 1

      at <anonymous> (/home/dmills/development/coder-modules/filebrowser/main.test.ts:33:27)
✗ filebrowser > runs with default [546.26ms]
46 |       agent_id: "foo",
47 |       database_path: ".config/filebrowser.db",
48 |     });
49 |     const output = await executeScriptInContainer(state, "debian:bookworm-slim");
50 |     expect(output.exitCode).toBe(0);
51 |     expect(output.stdout).toEqual([
                               ^
error: expect(received).toEqual(expected)

  [
+   ""
-   "\u001B[0;1mInstalling filebrowser ",
-   "",
-   "🥳 Installation complete! ",
-   "",
-   "👷 Starting filebrowser in background... ",
-   "",
-   "📂 Serving /root at http://localhost:13339 ",
-   "",
-   "Running 'filebrowser --noauth --root /root --port 13339 -d .config/filebrowser.db --baseurl ' ",
-   "",
-   "📝 Logs at /tmp/filebrowser.log"
  ]

- Expected  - 11
+ Received  + 1
```

filebrowser/run.sh
```diff
@@ -21,10 +21,21 @@ if [ "${DB_PATH}" != "filebrowser.db" ]; then
   DB_FLAG=" -d ${DB_PATH}"
 fi
 
+# Check if filebrowser db exists
+if [ ! -f ${DB_PATH} ]; then
```
mafredri, line 25:
```suggestion
if [ -n "$DB_FLAG" ]; then
```
Ideally I'd want DB_FLAG to be an array, so it can handle spaces in DB_PATH, but for now let's at least improve this check to avoid syntax errors.

filebrowser/run.sh
```diff
@@ -21,10 +21,21 @@ if [ "${DB_PATH}" != "filebrowser.db" ]; then
...
+  filebrowser $DB_FLAG config init >> ${LOG_PATH} 2>&1
+  filebrowser $DB_FLAG users add admin "" --perm.admin=true --viewMode=mosaic >> ${LOG_PATH} 2>&1
+fi
+
+filebrowser $DB_FLAG config set --baseurl=${SERVER_BASE_PATH} --port=${PORT} --auth.method=noauth --root=$ROOT_DIR >> ${LOG_PATH} 2>&1
```
mafredri, line 30:
```suggestion
  filebrowser $DB_FLAG config init 2>&1 | tee -a "${LOG_PATH}"
  filebrowser $DB_FLAG users add admin "" --perm.admin=true --viewMode=mosaic 2>&1 | tee -a "${LOG_PATH}"
fi

filebrowser $DB_FLAG config set --baseurl=${SERVER_BASE_PATH} --port=${PORT} --auth.method=noauth --root=$ROOT_DIR 2>&1 | tee -a "${LOG_PATH}"
```
Just a suggestion, this will allow showing the output in the Coder UI as well as the log file (useful in case of error).

filebrowser/run.sh
```diff
@@ -21,10 +21,21 @@ if [ "${DB_PATH}" != "filebrowser.db" ]; then
...
+
+printf "👷 Starting filebrowser in background... \n\n"
+
+
 printf "📂 Serving $${ROOT_DIR} at http://localhost:${PORT} \n\n"
 
-printf "Running 'filebrowser --noauth --root $ROOT_DIR --port ${PORT}$${DB_FLAG} --baseurl ${SERVER_BASE_PATH}' \n\n"
+printf "Running 'filebrowser $DB_FLAG' \n\n"
```
mafredri, line 37:
```suggestion
printf "Running 'filebrowser %s'\n\n" "$DB_FLAG"
```
Since we're using printf, this is the typical way to include variable content. 😄

filebrowser/run.sh
```diff
@@ -21,10 +21,21 @@ if [ "${DB_PATH}" != "filebrowser.db" ]; then
...
 
-filebrowser --noauth --root $ROOT_DIR --port ${PORT}$${DB_FLAG} --baseurl ${SERVER_BASE_PATH} > ${LOG_PATH} 2>&1 &
+filebrowser $DB_FLAG >> ${LOG_PATH} 2>&1 &
```
mafredri, line 39: I want to make it clear that this issue _was not introduced by this PR_.

But I'll just mention this again; this will break if `DB_PATH` has spaces in it. A better way would be to initialize this as an array and use it in a way that handles spaces:

```bash
args=()
if [ "${DB_PATH}" != "filebrowser.db" ]; then
	args+=(-d "${DB_PATH}")
fi
```

Then use this as:

```bash
filebrowser "${args[@]}"
```

(I did not verify if `$$` vs `$` is required here due to terraform templates.)

This syntax works even in bash-3.2, so should be fine to use. (I opted for arrays as I see we already have a heavy dependency on Bash here.)

mafredri (commented):
Had some minor suggestions, but didn't spot anything terribly wrong that wasn't already. 😄

Also, since this is a Bash dependent script, we should use `[[` instead of `[`. The former has many benefits and no downsides compared to `[`.

matifali:
Hi @evilhamsterman, thank you for your contribution. Could you please address the review comments to merge this PR?

evilhamsterman:
@matifali Sorry I've had a very busy month. I should be able to get to it later this week

filebrowser/run.sh
evilhamsterman: This needs to be DB_PATH here. It's checking if the file exists so it can initialize it if it doesn't.

filebrowser/run.sh
evilhamsterman: I thought of doing that myself, but decided to stick with the established convention

filebrowser/run.sh
evilhamsterman: Again following convention, changes made

filebrowser/run.sh
evilhamsterman: TBH There's a lot of places that it could break because of spaces in the path, the LOG_PATH and SERVER_BASE_PATH could also cause problems.

I do think I found a bit easier and more readable way to handle it. Filebrowser supports the FB_DATABASE environment variable. So I just set that at the top inside quotes and then it doesn't matter, it works with or without spaces.

evilhamsterman:
I applied some of the suggestions above, I also resolved the issue of spaces in the DB path by using the FB_DATABASE environment variable instead. Let me know if there's anything else

phorcys420:
@evilhamsterman the reason our tests are failing (even on main) is due to the fact the script is non-strict and ignores any `stderr` output, so the logs are correct even though everything is broken, basically meaning the tests are useless in this context.

see this example
image

phorcys420:
@coder

phorcys420 (approved):
I have confirmed that this works by creating a [template](https://dev.coder.com/templates/coder/coder-module-400) on our internal instance.

image

phorcys420:
@matifali could you force this merge?

matifali:
Force merged.

evilhamsterman:
Thanks for the help!

PR #402, opened 2025-02-13
feat(vscode-web): allow pinning vscode-web binary to a specific commit ID

rogmanster:
Adds support for specifying a commit ID to pin the vscode-web binary version in the module.

### DIFF
--- a/vscode-web/README.md
+++ b/vscode-web/README.md
@@ -15,7 +15,7 @@ Automatically install [Visual Studio Code Server](https://code.visualstudio.com/
 module "vscode-web" {
   count          = data.coder_workspace.me.start_count
   source         = "registry.coder.com/modules/vscode-web/coder"
-  version        = "1.0.29"
+  version        = "1.0.30"
   agent_id       = coder_agent.example.id
   accept_license = true
 }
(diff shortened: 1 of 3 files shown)


vscode-web/main.tf
```diff
@@ -59,6 +59,12 @@ variable "install_prefix" {
   default     = "/tmp/vscode-web"
 }
 
+variable "vscode_web_commit_id" {
```
matifali, line 62:
```suggestion
variable "commit_id" {
```
We can shorten the name.

vscode-web/README.md
matifali, line 1: Can you add an example with a specified `commot_id`? Also describe how to get the commit id.

matifali (approved):

matifali:
@rogmanster can you follow up with the suggestions?

vscode-web/README.md
rogmanster: Thanks for the suggestions @matifali! Since the PR is merged, I follow-up with a new PR to shorten variable name and add an example in the README?

matifali: Yes sounds great. Thank you 😸

matifali: I will also check if auto-merge is enabled before hitting approve.
I will wait until then to cut the release.
Also, we can revert this PR and add all changes in a new one to keep the changelog simple.

PR #403, opened 2025-02-14
Revert "feat(vscode-web): allow pinning vscode-web binary to a specific commit ID"

matifali:
Reverts coder/modules#402

We are redoing it with suggested changes.


rogmanster (approved):
LGTM

PR #405, opened 2025-02-14
feat(vscode-web): allow pinning vscode-web version to a specific commit ID

rogmanster:
Adds support for specifying a commit ID to pin the vscode-web version in the module with added example in README.


matifali (approved):
LGTM. Thanks @rogmanster

PR #406, opened 2025-02-17
feat(jupyterlab): added support for uv as installer

ffais:
Support installation of jupyterlab with UV.


matifali (approved):
Thanks for your contribution @ffais.

PR #407, opened 2025-02-18
fix(jupyterlab): don't use `-q` option

mats16:
On Ubuntu 22.04, pipx (via `apt install`) does not have the `-q` option. 
The description that depends on the pipx version should be reduced.


matifali (approved):
Thanks. We probbaly did not test it for 22.04

matifali (approved):

matifali:
Hi @mats16, can you please take a look at the merge conflict?

matifali:
We do not use `-q` anymore after the refactoring in #406 

https://github.com/coder/modules/blob/main/jupyterlab/run.sh#L40

PR #411, opened 2025-02-26
Use VSCode Code CLI instead of VSCode code-server to run VSCode Web

Ricky-Hao:
Hi,
The current VSCode Web module directly uses `code-server` to start a VSCode Web instance.
However, I noticed that the current VSCode Web only stores user credentials (e.g., GitHub authentication) in in-memory storage.

After investigating the VSCode source code, I discovered that VSCode Web requires a key from the `/mint-key` endpoint.
In Coder's `code-server`, the `/mint-key` endpoint is implemented.
However, in the official VSCode version, the `/mint-key` functionality is only available in the VSCode CLI (`code` binary).

To enable credential persistence, we need to use `code serve-web` to start a VSCode Web instance instead of the current `code-server serve-local`.

In this PR, I updated the process to use `code serve-web` to start a VSCode Web instance.
Most of the features from the old `code-server` are supported, except for the following two:

- `OFFLINE`
- `USE_CACHE`

I did not remove these variables from the `.tf` files to maintain compatibility.

### DIFF
--- a/vscode-web/run.sh
+++ b/vscode-web/run.sh
@@ -5,21 +5,19 @@ EXTENSIONS=("${EXTENSIONS}")
 VSCODE_CLI="${INSTALL_PREFIX}/code"
 
 # Set extension directory
-EXTENSION_ARG=""
 if [ -n "${EXTENSIONS_DIR}" ]; then
-  EXTENSION_ARG="--extensions-dir=${EXTENSIONS_DIR}"
+  EXTENSIONS_DIR="--extensions-dir=${EXTENSIONS_DIR}"
 fi
 
 # Set extension directory
-SERVER_BASE_PATH_ARG=""
 if [ -n "${SERVER_BASE_PATH}" ]; then
-  SERVER_BASE_PATH_ARG="--server-base-path=${SERVER_BASE_PATH}"
+  SERVER_BASE_PATH="--server-base-path=${SERVER_BASE_PATH}"
 fi
 
 run_vscode_web() {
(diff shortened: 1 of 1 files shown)


bcpeinhardt:
@matifali this would be a good opportunity to use the new individual version tag structure when releasing.

matifali:
Yes if this. lands, it should be go as 2.0.0 for vscode-web. @bcpeinhardt can you take care of this?

matifali:
Hi @Ricky-Hao, Sorry for not coming to this earlier. We were sorting versions for the modules, and as this module changes stuff significantly while Aldo breaks a few existing =g use cases, We needed that to land this in.

Can you update the `README.md` to reflect the changes? Specifically the ones which are o more supported.

matifali:
I would also like @michaelbrewer to review as he has contributed to this module in past.

matifali:
> Most of the features from the old code-server are supported, except for the following two:
> - `OFFLINE`
> - `USE_CACHE`
>
> I did not remove these variables from the .tf files to maintain compatibility.

Hi @Ricky-Hao, why can we not support these two features with VS Code's `code serve-web` CLI?

Ricky-Hao:
> Hi @Ricky-Hao, why can we not support these two features with VS Code's `code serve-web` CLI?
Hi @matifali, sorry for the delayed response.
Since I’m not using these two features in my workspace, I’m not sure what they are supposed to do.

matifali:
I think the difference here is that `code serve-web` also manages downloading the `code-server` binary, and we do not know if we can make it use an existing binary. IF we can somehow instruct it to use an already downloaded `code-server` binary, it should work with `OFFLINE` and `USE_CACHED` modes.

matifali:
More context in #469

PR #415, opened 2025-03-13
feat(code-server): link binary to PATH

mafredri:
Noticed we don't place `code-server` in path, so made a quick fix.


matifali (approved):
Thanks 👍. 
We should try to do this uniformaly for all our modules.

phorcys420:
went ahead and fixed the version mismatch so we can merge this

PR #430, opened 2025-04-15
feat(code-server): make `open_in` configurable

valeneiko:
`open_in` option for Coder App was introduced in `2.1.0`. It would be nice to expose it in `code-server` so that we can configure `code-server` to open in new tab instead of new window.

### DIFF
--- a/code-server/main.tf
+++ b/code-server/main.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     coder = {
       source  = "coder/coder"
-      version = ">= 0.17"
+      version = ">= 2.1"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


owre (approved):

LorenzHenk (approved):

matifali:
Hi @valeneiko, Thanks for the contribution. It may be a breaking change for users with a hard coder and a provider version of less than 2.1.0 in their templates.

We should bump the minor module version to prevent the module from auto-updating.

valeneiko:
@matifali I updated version in README. Is that what you meant?

code-server/main.tf
```diff
@@ -122,6 +122,16 @@ variable "subdomain" {
...
 
+variable "open_in" {
+  type        = string
+  description = <<-EOT
+    Determines where the app will be opened. Valid values are `"tab"` and `"slim-window" (default)`.
+    `"tab"` opens in a new tab in the same browser window.
+    `"slim-window"` opens a new browser window without navigation controls.
+  EOT
+  default     = "slim-window"
+}
```
matifali, line 133: Can you add a [`validation`](https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules) block to check if the provided value is one of the valid options?

valeneiko: done

matifali (approved):

PR #432, opened 2025-04-16
feat(claude-code): use tmux instead of screen 

bcpeinhardt:


claude-code/README.md
matifali, line 1: Can we bump the version to 1.1.0? in README.md?

bcpeinhardt: Ah good catch thank you!

bpmct (approved):

PR #435, opened 2025-04-22
fix(vault-jwt): store vault token for use in vault jwt module

moo-im-a-cow:
currently, the vault jwt module generates a vault token, but it doesnt actually store it for future vault commands to use.

even with the module configured, running `vault token lookup` fails, because the login command in the module just prints the token but doesnt keep it to be used by future calls to vault

added `-field=token` to the write command, which makes it print only the token itself to stdout, not the rest of the info about the token
then pipe that into `vault login -`, which reads from stdin, and does the process of actually storing the token for future use.

(compared to the other vault coder modules which both call `vault login`, but this one doesnt, but it should.)


matifali (approved):
Thanks for contribution.

PR #436, opened 2025-04-22
feat(vault-jwt): allow specifying the vault jwt token directly

moo-im-a-cow:
this PR allows you to pass the desired jwt token to the vault-jwt module as a variable, for cases when you want to use a different jwt token to the one used to authenticate to vault

if not specified, it defaults to `data.coder_workspace_owner.me.oidc_access_token`.


vault-jwt/main.tf
```diff
@@ -20,6 +20,13 @@ variable "vault_addr" {
   description = "The address of the Vault server."
 }
 
+variable "vault_jwt_token" {
+  type        = string
+  description = "The JWT token used for authentication with Vault."
+  default     = null
+  sensitive   = true
+}
+
```
matifali, line 29: Can you add an example use case on how you would provide that token to the module? Are you fetching it externally through some API or another provider?

matifali: > do you want the examples added to the commit in some way?
i'm unsure how documentation works here, but i'll do whatever is needed

We usually add example Terraform snippets in the `README.md` to guide users on possible ways the module can be used.

matifali: Thanks. I would appreciate it if we could also add a vault policy that can use this new jwt token, scoped to workspace metadata.

matifali (commented):
Not a vault expert at all. But as far as I understand the the workspace scoped secrets need to be pre created in vault. How would one do that. 
In other words, I agree with use case to provide workspace only secrets access but is concerned if it makes the config difficult on the vault side. How would one provision these secrets? For what type of secrets this is most useful for.

Thanks.

matifali:
Hi @moo-im-a-cow can you run `bun fmt`?

matifali:
@mafredri, what is the best way to pass this prettier validation check? AFAIR, we did this as the plugin we were using was not able to format `hcl` blocks, and only works for `tf` blocks.

DevelopmentCats:
I have changed the syntax and code block in the readme for the Vault Policy example.

I will do one last test of this module using these new blocks to make sure it works properly before Accepting

DevelopmentCats:
I'm going to pull this locally to test it before approving this but everything is looking good now.

vault-jwt/README.md
```diff
@@ -79,3 +80,105 @@ module "vault" {
...
+  count     = data.coder_workspace.me.start_count
+  algorithm = "RS256"
+  # `openssl genrsa -out key.pem 4096` and `openssl rsa -in key.pem -pubout > pub.pem` to generate keys
+  key = file("key.pem")
+  claims_json = jsonencode({
+    iss = "https://code.example.com"
+    sub = "${data.coder_workspace.me.id}"
+    aud = "https://vault.example.com"
+    iat = provider::time::rfc3339_parse(plantimestamp()).unix
+    # exp = timeadd(timestamp(), 3600)
```
mafredri, line 112: This should either be removed or have an associated comment (uncomment to ...).

vault-jwt/README.md
```diff
@@ -79,3 +80,105 @@ module "vault" {
...
+    "provisioner_arch": "provisioner_arch",
+    "provisioner_os": "provisioner_os",
+    "sub": "sub",
+    "template": "template",
+    "template_name": "template_name",
+    "template_version": "template_version",
+    "workspace": "workspace",
+    "workspace_name": "workspace_name",
+    "workspace_id": "workspace_id"
+}
```
mafredri, line 168:
```suggestion
  }
```

vault-jwt/README.md
```diff
@@ -79,3 +80,105 @@ module "vault" {
...
+}
+EOF
+```
+
+#### example workspace access vault policy
```
mafredri, line 173:
```suggestion
#### Example workspace access Vault policy
```

vault-jwt/README.md
```diff
@@ -79,3 +80,105 @@ module "vault" {
...
+  source          = "registry.coder.com/modules/vault-jwt/coder"
+  version         = "1.0.20"
+  agent_id        = coder_agent.example.id
+  vault_addr      = "https://vault.example.com"
+  vault_jwt_role  = "coder" # The Vault role to use for authentication
+  vault_jwt_token = jwt_signed_token.vault[0].token
+}
+```
+
+#### example vault jwt role
```
mafredri, line 144:
```suggestion
#### Example Vault JWT role
```

vault-jwt/README.md
```diff
@@ -79,3 +80,105 @@ module "vault" {
   vault_cli_version = "1.17.5"
 }
 ```
+
+### use a custom jwt token
```
mafredri, line 84:
```suggestion
### Use a custom JWT token
```

mafredri (commented):
Just a couple minor nits, otherwise LGTM 👍🏻

vault-jwt/README.md
```diff
@@ -10,16 +10,17 @@ tags: [helper, integration, vault, jwt, oidc]
...
 module "vault" {
-  count          = data.coder_workspace.me.start_count
-  source         = "registry.coder.com/modules/vault-jwt/coder"
-  version        = "1.0.21"
-  agent_id       = coder_agent.example.id
-  vault_addr     = "https://vault.example.com"
-  vault_jwt_role = "coder" # The Vault role to use for authentication
+  count           = data.coder_workspace.me.start_count
+  source          = "registry.coder.com/modules/vault-jwt/coder"
+  version         = "1.0.21"
```
matifali, line 19:
```suggestion
  version         = "1.1.0"
```
Bump version to `1.1.0`

vault-jwt/README.md
moo-im-a-cow: assuming you only meant the exp field, i've fixed it up and added a comment describing it and the pro/con of uncommenting it (making the token expire)

matifali:
Should be available in version `1.1.0`
Thanks a lot for the contribution.

PR #441, opened 2025-04-22
ci: deploy dev registry on tag pushes

matifali:
This pull request updates the GitHub Actions workflow to enhance deployment flexibility by allowing it to trigger on specific tag patterns in addition to pushes to the `main` branch.

Workflow trigger update:

* [`.github/workflows/deploy-registry.yaml`](diffhunk://#diff-fbf55ed2cd70e3380da25686178b863d15e0dcd8b5c8f4115ea33fae51bbd883R7-R8): Added a `tags` condition under the `push` event to trigger the workflow for tags matching the pattern `release/*/v*` (e.g., `release/module-name/v1.0.0`).

### DIFF
--- a/.github/workflows/deploy-registry.yaml
+++ b/.github/workflows/deploy-registry.yaml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - "release/*/v*" # Matches tags like release/module-name/v1.0.0
 
 jobs:
   deploy:


bcpeinhardt (approved):

bcpeinhardt:
Nit: I think we'll have quite a few started and cancelled builds, where a tag is pushed just after a merge.

matifali:
> Nit: I think we'll have quite a few started and cancelled builds, where a tag is pushed just after a merge.

We do not have any [concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-using-concurrency-and-the-default-behavior) settings that will allow parallel builds. 

How do we fetch tags during the build process? I don't want the latest pushed tag to an older commit to somehow exclude all later tags from the latest commit. 

Without proper concurrency settings, module tags pushed to older commits can cause build issues:

```
A --- B --- C --- D  ← Main branch 
      ↑     ↑     ↑
      |     |     |
module-a/v1.0.21  module-b/v1.2.0
            |
            ↓
      module-a/v1.1.0
      (pushed later to an older commit)
```

Would this new tag (`module-a/v1.1.0`) pushed to an older commit cause the newer commit tag (`module-b/v1.2.0`) to be excluded from the latest build?

PR #442, opened 2025-04-23
fix(claude-code): add `dangerously-skip-permissions`

kylecarbs:


bcpeinhardt:
I changed the PR to use the `-c` flag again since we confirmed it does work.

PR #447, opened 2025-04-23
fix(jupyterlab): fix path to jupyter-lab if already installed

cirego:
If the `jupyter-lab` binary already exists in $PATH, as found by `command -v jupyter-lab`, then we skip installing using `pipx` or `uv`.

By skipping the install path, we leave `${JUPYTERPATH}` unset, which then leads to a broken module unless `jupyter-lab` just so happens to be installed at `/jupyter-lab`.

Fix this by calculating the full path to the binary, either from the install locations of `pipx`, `uv` or, in the case that it was preinstalled, from the preinstalled location.

I did not test this. I would have added a test but the current tests don't run because
they're too slow.

### DIFF
--- a/jupyterlab/README.md
+++ b/jupyterlab/README.md
@@ -17,7 +17,7 @@ A module that adds JupyterLab in your Coder template.
 module "jupyterlab" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/modules/jupyterlab/coder"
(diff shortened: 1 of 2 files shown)


matifali (approved):
Thanks for the fix. Can you please bump the patch version in README.md examples?

It's a manual thing for now, we will soon automate it.

PR #449, opened 2025-04-25
feat(code-server): Machine Settings option

djarbz:
This PR resolves https://github.com/coder/modules/issues/443.

The goal is to provide a method for template owners to configure default `Machine Settings` that can be overridden by developers via `User Settings` and repositories via `Workspace Settings`.
This option allows template owners to push new settings options with a template release that would not be ignored because the setting file already exists.

Eventually, I would imagine the current `settings` option will be depreciated in favor of this option.
Alternatively, we could hijack that option to operate on the machine level settings before merging this PR.

### DIFF
--- a/code-server/main.tf
+++ b/code-server/main.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     coder = {
       source  = "coder/coder"
-      version = ">= 2.1"
+      version = ">= 0.17"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


matifali (changes requested):
Please do not remove existing functionality
And bump minor version.

djarbz:
Looks like my fetch sync didn't work properly the first go-around.
Should be good now.

matifali:
More context in #469

PR #452, opened 2025-04-28
feat(jetbrains-gateway): add arm64 support

diranged:
Provide support for running Jetbrains Gateway connections to both ARM64 and AMD64 hosts.

### DIFF
--- a/jetbrains-gateway/README.md
+++ b/jetbrains-gateway/README.md
@@ -18,7 +18,7 @@ Consult the [JetBrains documentation](https://www.jetbrains.com/help/idea/prereq
 module "jetbrains_gateway" {
   count          = data.coder_workspace.me.start_count
   source         = "registry.coder.com/modules/jetbrains-gateway/coder"
-  version        = "1.0.28"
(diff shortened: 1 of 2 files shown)


phorcys420:
FYI you can bump via `./update_version.sh`

diranged:
@phorcys420 So I bumped it by a patch version - `1.0.28 -> 1.0.29` only because it didn't seem like any previous PRs had bumped the minor version. Is that what you were looking for?

phorcys420:
yes @diranged !
we don't do proper semver for modules at the moment so this is how it's done currently.

phorcys420 (approved):
LGTM
thanks a lot for your contribution!

matifali:
@phorcys420, we do follow semver now. We do not have strict version checking to enforce this, so the maintainers are responsible for ensuring the correct version bump.

This was supposed to be a minor version bump, as I requested in https://github.com/coder/modules/pull/452#pullrequestreview-2813148937

phorcys420:
ah alright my bad thanks for letting me know!

PR #466, opened 2025-05-09
feat(vault-token): make supplying a vault token optional

moo-im-a-cow:
makes the vault token variable optional, 
meaning that template creators can use the vault token module just to install vault cli, without it configuring a token environment variable.
the template creator can then log into vault with their own script, 
or they can ask the user to manually run `vault login` command

potential future change: refactor the vault-jwt and the vault-github modules in this repo to call this module for installing vault, and their scripts only need to call the relevant login command, they dont need to download and install vault. makes maintaining easier if you have just one install script.

### DIFF
--- a/vault-token/main.test.ts
+++ b/vault-token/main.test.ts
@@ -7,6 +7,5 @@ describe("vault-token", async () => {
   testRequiredVariables(import.meta.dir, {
     agent_id: "foo",
     vault_addr: "foo",
-    vault_token: "foo",
   });
 });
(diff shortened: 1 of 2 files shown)


moo-im-a-cow:
my particular usecase: 
I switched from having a jwt generated in the coder template, 
instead i now have a k8s operator that generates a new vault token (https://github.com/hashicorp/vault-secrets-operator/pull/1062), and saves it to a k8s secret. 
that token is short lived and is rotated regularly, the k8s secret always has a valid freshly generated vault token

the vault token secret is mounted as a file to the workspace container, 
and is symlinked over the `~/.vault-token` file
therefore, no authentication to vault ever occurs inside the container, 

providing the token via envvar would result in the token expiring and not being updated with the secret, 
and i didnt want to create my own new vault coder module

matifali:
More context in #469

PR #471, opened 2025-05-09
chore: update README.md

matifali:
image

I think we should plan to archive this repo by the end of next week after we have tested the new registry deployment.


README.md
```diff
@@ -1,3 +1,6 @@
+> [!CAUTION]
+> We are no more accepting new contributions to this repo. We have moved all modules to https://github.com/coder/registry repo. Please see https://github.com/coder/modules/discussions/469 for more details.
```
bcpeinhardt: > no more accepting

should probably be "no longer accepting"

bcpeinhardt (changes requested):

README.md
matifali: udpated

bcpeinhardt (approved):

